PR #1, opened 2023-09-28
Async Aggregators

janine9vn:


async-aggregators/src/utils/apply_ascii_art.py
```diff
@@ -0,0 +1,29 @@
...
+
+    :param args: a dictionary of arguments
+    :return: QPixmap of the image
+    """
+    try:
+        image_dir_path = Path(Path(__file__).parent.parent, "images")
+        input_img, coordinates = prepare_input(Path(str(args.get("image_to_edit"))))
+        output_img_path = Path(image_dir_path, "ascii_output.png")
+        ascii_file_path = Path(image_dir_path, "ascii.txt")
+        generate_ascii_file(input_img, ascii_file_path, 2)
```
camcaswell, line 22: Is there a purpose to saving the text to a file before immediately reopening it in `seed_secret` and `ascii_to_image`? It would probably be better to keep the text in memory and save the four file IO operations.
It looks like PIL has a way to [convert to a QImage](https://pillow.readthedocs.io/en/stable/reference/ImageQt.html), which `QPixmap` can take.

camcaswell: I would expect a function named `ascii_to_image` to just return an image object. If its main purpose is a side-effect like writing a file, I'd prefer that to be reflected in the name, e.g. `write_ascii_png` or something.

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+    size = min(width, height)
+
+    # Calculate the coordinates for cropping
+    left = (width - size) // 2
+    top = (height - size) // 2
+    right = (width + size) // 2
+    bottom = (height + size) // 2
+
+    # Crop the image to a square
+    return img.crop((left, top, right, bottom)), [left, top, right, bottom]
```
camcaswell, line 29: This function's job confuses me a little bit. As far as I can tell those coordinates are only used to calculate the width and height of the cropped section, but those are properties of the `Image` object it's returning anyway.
Clarifying the purpose/type of the function would also allow giving it a more descriptive name, e.g. `square_crop`.

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+
+    :param img: input image
+    :return: average luminance
+    """
+    np_img = np.array(img)
+    w, h = np_img.shape
+    return np.average(np_img.reshape(w * h))
+
+
+def img_to_ascii(img: Image.Image, dens: int) -> List[str]:
```
camcaswell, line 44: Since 3.9, you can use the built-in types here (note the capitalization of the typehint).
```suggestion
def img_to_ascii(img: Image.Image, dens: int) -> list[str]:
```

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+    gs_img_w, gs_img_h = gs_img.size[0], gs_img.size[1]
+    cols = int(gs_img_w // 10)  # 1/10th of the cropped square
+    scale = 0.39  # Aspect ratio of monospace characters
+
+    w = gs_img_w / cols  # tile width
+    h = w / scale  # tile height
+    rows = int(gs_img_h / h)  # Number of rows of text
+
+    # Generate ascii art
+    ascii_ = []
```
camcaswell, line 70: When initializing an empty list like this, it's good to typehint the values that will be added to it later.
```suggestion
    ascii_: list[str] = []
```

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+    # The amount of text to white space in each of these character determines the greyscale value
+    greyscale_strs = [
+        "@%#*+=-:. ",  # 10 characters,
+        "@%W*hdwOLUzurt|1i}?+>,' ",  # 24 characters,
+        "@B$%8&WM#*oahkbdpqwmZO0QLCJUYXzcvunxrjft/\\|()1lIi!{}[]?-_+~<>;:,^`'. ",  # 70 characters,
+    ]
+
+    # Convert to greyscale (L stands for luminance)
+    gs_img = img.convert("L")
+    gs_img_w, gs_img_h = gs_img.size[0], gs_img.size[1]
```
camcaswell, line 61:
```suggestion
    gs_img_w, gs_img_h = gs_img.size
```

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+        y1 = int(r * h)
+        y2 = gs_img_h if r == rows - 1 else int((r + 1) * h)
+        ascii_.append("")
+
+        for c in range(cols):
+            x1 = int(c * w)
+            x2 = gs_img_w if c == cols - 1 else int((c + 1) * w)
+
+            # Find average luminance of the cropped tile
+            avg = int(avg_greyscale(gs_img.crop((x1, y1, x2, y2))))
```
camcaswell, line 81: I suspect it would be better to convert the image to a numpy array once then take the average of slices of that array, rather than creating a new image each time and turning that into an array.

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+def img_to_ascii(img: Image.Image, dens: int) -> List[str]:
+    """
+    Given Image and dims (rows, cols) returns an m*n list of Images
+
+    :param img: input image file
+    :param dens: greyscale density level
+    :return: ascii text generated from input image
+    """
+    # The amount of text to white space in each of these character determines the greyscale value
+    greyscale_strs = [
```
camcaswell, line 53: An additional comment to explain the difference between these character sets could be useful.

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+
+        for c in range(cols):
+            x1 = int(c * w)
+            x2 = gs_img_w if c == cols - 1 else int((c + 1) * w)
+
+            # Find average luminance of the cropped tile
+            avg = int(avg_greyscale(gs_img.crop((x1, y1, x2, y2))))
+
+            # Find corresponding ascii character by greyscale value and append to the row
+            i = int((avg * (len(greyscale_strs[dens]) - 1)) / 255)
```
camcaswell, line 84: This could be split up to make it clearer
```py
characters = greyscale_strs[dens]
...
    grey_value = avg / 255
    char_idx = int(grey_value * (len(characters) - 1))
    ascii_[r] += char_set[characters]
```

async-aggregators/lib/hidden_in_ascii/hidden_in_ascii.py
```diff
@@ -0,0 +1,183 @@
...
+    # Convert to binary representation if chosen 'insane mode'
+    if binary_mode:
+        secret = "".join(map(bin, bytearray(secret, "utf8")))
+
+    line_length = len(lines[random_line])
+    secret_length = len(secret)
+
+    position = random.randint(0, line_length - secret_length - 1)
+    lines[random_line] = (
+        "".join([lines[random_line][0:position], secret, lines[random_line][position + secret_length:]])
```
camcaswell, line 141:
```suggestion
        old_line[:position] + secret + old_line[position+secret_length:]
```

async-aggregators/lib/motions/motions.py
```diff
@@ -0,0 +1,236 @@
...
+
+class OFF_CANVAS_FILL(Enum):
+    """Methods used to fill pixels from outside the image canvas"""
+
+    WRAP = auto()
+    MIRROR = auto()
+    EDGE = auto()
+
+
+class Effect:
```
camcaswell, line 33: Usually classes in python aren't used for purely taxonomical purposes. I think this one is defensible since it's basically an enum of functions.
One simple alternative would be to just move these functions into their own module.
In 3.11 there's a new [`enum.member`](https://docs.python.org/3/library/enum.html#enum.member) decorator that you could use to explicitly create an enumeration (I realize this project uses 3.10).

async-aggregators/src/control_panel.py
```diff
@@ -0,0 +1,225 @@
...
+
+class ControlPanel(QWidget):
+    """Control Panel"""
+
+    # Define a new signal at the top of the class
+    sliderValueChanged = pyqtSignal(str, int)
+    comboBoxesSwapped = pyqtSignal(str, str)
+    ascii = pyqtSignal()
+
+    def __init__(self, title: str, widget_info: dict):
```
camcaswell, line 56: `widget_info` should be split up into keyword arguments. As is, it's hard to tell what info is expected by the class.

Edit: Coming back to this as I've seen the same pattern in multiple other places. An argument called "args" that holds other arguments serves only to obfsucate what data is being passed to the function. It seems like the purpose is to implement keyword arguments? Just use keyword arguments.

async-aggregators/src/level.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Get the filters for the level
+
+        :return: list of filters
+        """
+        # This can be extended to provide filters dynamically based on the level
+
+        icons_dir_path = Path(Path(__file__).parent, "icons")
+        image_dir_path = Path(Path(__file__).parent, "images")
+        filters = [
```
camcaswell, line 73: Why is each tuple a singleton in a list? Why is the list named `filters` a list of `FilterList`s and not a list of `FilterItem`s?

async-aggregators/src/level.py
```diff
@@ -0,0 +1,265 @@
...
+from lib.motions.motions import MotionTransformer
+from src.control_panel import ControlPanel
+
+FilterItem = Tuple[Path, ControlPanel, Dict[str, Path | str | None | int]]
+FilterList = List[FilterItem]
+letters = [chr(i) for i in range(65, 65 + 21)]
+
+
+class Level:
+    """Class to represent a level in the game."""
```
camcaswell, line 16: This class does not actually represent a single level, it seems to represent a whole game. The problem is not just that the docstring is misleading though; there probably *should* be a decoupling of the level information from the gamestate. The game knows which level the player is currently on, and each level is a collection of static information.

Each level should be defined once and only know about the things that pertain to that one level. As is, `get_filters` is re-initializing all the level data every time it's called.

Something more like this:
```py
LEVELS = [
    Level(title="Ishihara", secret="42", image_path=Path(...), etc=etc),
    Level(title="Double Exposure", secret="snake_case", image_path=Path(...), etc=etc),
    ...
]

class Game:
    def load_level(self):
        display(LEVELS[self.level_number])
```

async-aggregators/src/level.py
```diff
@@ -0,0 +1,265 @@
...
+                        "secret_code": "codes",
+                        "MotionTransformer": MotionTransformer(
+                            Image.open(image_dir_path / "snake.jpg"),
+                            "codes"),
+                    },
+                ),
+            ],
+        ]
+
+        if 0 <= self.level_number - 1 < len(filters):
```
camcaswell, line 250: The other comment about the general structure of this class is more important, but at least it should check whether the level index is valid *before* going through all of that.

async-aggregators/src/level.py
```diff
@@ -0,0 +1,265 @@
...
+
+        :return: list of filters
+        """
+        # This can be extended to provide filters dynamically based on the level
+
+        icons_dir_path = Path(Path(__file__).parent, "icons")
+        image_dir_path = Path(Path(__file__).parent, "images")
+        filters = [
+            [
+                (
```
camcaswell, line 75: RE the comment about making a `Level` class to represent one level:
Implementing it that way would also make this easier to read, since these things would be separate arguments to `Level.__init__`.

async-aggregators/lib/motions/motions.py
```diff
@@ -0,0 +1,236 @@
...
+                1,
+            )
+            + 0.0000001
+        )
+        xmesh = (xmesh - width / 2) * new_distance / normalized_distance + width / 2
+        ymesh = (ymesh - height / 2) * new_distance / normalized_distance + height / 2
+        return xmesh, ymesh
+
+
+def explode(
```
camcaswell, line 135: Mix-up between this and `Effect.explode`?

async-aggregators/src/utils/apply_motions.py
```diff
@@ -0,0 +1,32 @@
+from PyQt6.QtGui import QImage, QPixmap
+
+from lib.motions.motions import MotionTransformer
+
+
+def apply_motion(args: dict) -> QPixmap:
```
camcaswell, line 6: Why not make this a method of `MotionTransformer`?

async-aggregators/lib/motions/motions.py
```diff
@@ -0,0 +1,236 @@
...
+            (2 * xmesh / width - 1) ** 2 + (2 * ymesh / height - 1) ** 2
+        ) ** 0.5
+        new_distance = normalized_distance / (
+            np.maximum(
+                (1.5 - normalized_distance) * 3 * (abs(magnitude) + 0.000001) ** 0.8
+                + 0.2,
+                1,
+            )
+            + 0.0000001
+        )
```
camcaswell, line 129: The math and magic numbers could use some explanation.

async-aggregators/lib/pixelate_and_swap/pixelate_and_swap.py
```diff
@@ -0,0 +1,76 @@
...
+    # Assign all pixels to nearest cluster centers
+    labels = kmeans.predict(pixels)
+    new_pixels = np.array([list(cluster_centers[label]) for label in labels], dtype=np.uint8)
+
+    # Reshape to original image shape
+    new_image_data = new_pixels.reshape(data.shape)
+
+    # Convert back to Image and return
+    new_image = Image.fromarray(new_image_data)
+    return new_image
```
camcaswell, line 45: This is a small thing, but I would usually skip the name definition in cases like this. It saves a line and coming up with a name.
```suggestion
    return Image.fromarray(new_image_data)
```

async-aggregators/src/filter.py
```diff
@@ -0,0 +1,203 @@
...
+                args_for_filter["B"] = value
+            if key == "image_to_edit":
+                args_for_filter["image_to_edit"] = value
+            if key == "image_label_w":
+                args_for_filter["image_label_w"] = value
+            if key == "image_label_h":
+                args_for_filter["image_label_h"] = value
+        return apply_unmask_reverse_ishihara(
+            args_for_filter
+        )
```
camcaswell, line 168: I like splitting up long function calls onto multiple lines, but this can fit easily into one.
```suggestion
        return apply_unmask_reverse_ishihara(args_for_filter)
```

async-aggregators/src/dock.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Define a new signal at the top of the class
+    controlValueChanged = pyqtSignal(str, int)
+
+    def __init__(self, level: Level, image_label: QLabel, image_display: QWidget,
+                 update_secret_code_callback: Callable) -> None:
+        super().__init__()
+
+        self.level = level
+        self._update_secret_code = update_secret_code_callback
```
camcaswell, line 27: This feels a little weird to me. This is essentially getting a reference to the window through the bound method. I think it would be better to explicitly pass the window object itself and either write a function in this class that can use it, or just call `window._update_secret_code()`.

Which option to go with should depend on whether the functionality of updating the secret code should be determined by the window or the dock.

async-aggregators/lib/xor_mask/xor_mask.py
```diff
@@ -0,0 +1,123 @@
...
+                    image2_array[i][j] = 0
+            else:
+                # If the original pixel is black, randomly choose one of the image to be black pixel
+                # and the other image to be white pixel
+                if np.random.randint(2) == 0:
+                    image1_array[i][j] = 0
+                    image2_array[i][j] = 255
+                else:
+                    image1_array[i][j] = 255
+                    image2_array[i][j] = 0
```
camcaswell, line 85:
```suggestion
            original_pixel = int(original_array[i][j] >= 128)
            pixel_1 = np.random.randint(2)
            pixel_2 = original_pixel != pixel_1
            image1_array[i][j] = 255 * pixel_1
            image2_array[i][j] = 255 * pixel_2
```

async-aggregators/lib/xor_mask/xor_mask.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # Convert numpy arrays back to images
+    image1 = Image.fromarray(image1_array)
+    image2 = Image.fromarray(image2_array)
+
+    # Save the two new images
+    output_directory = Path("images/xor_mask")
+    output_directory.mkdir(parents=True, exist_ok=True)
+    image1.save(Path(f"images/xor_mask/{alphanumeric}_1.png"))
```
camcaswell, line 94: Would be good to use `output_directory` here since it was just defined.

async-aggregators/src/utils/apply_color_swap.py
```diff
@@ -0,0 +1,49 @@
...
+    :return: QPixmap
+    """
+    colors = {
+        "Rust": (164, 43, 17),
+        "Chocolate": (229, 95, 24),
+        "Flamenco": (236, 135, 67),
+        "Casablanca": (238, 171, 98),
+        "Buff": (242, 217, 143)
+    }
+    img_path = str(image)
```
camcaswell, line 29: The typehint says `image` is a tuple, the docstring says it's a path. After looking at where this function is used, it's actually a stringified path or the int `0`.

async-aggregators/src/utils/lru_cache_cv2.py
```diff
@@ -0,0 +1,41 @@
+from collections import OrderedDict
+
+import numpy as np
+
+
+class LRUCacheCV2:
```
camcaswell, line 6: This and `LRUCachePIL` do the same thing. The only real difference is the typehint, but there are [much better ways](https://docs.python.org/3.10/library/typing.html#typing.Generic) to handle that.
Furthermore, why not just use the [LRU cache implementation](https://docs.python.org/3.10/library/functools.html?highlight=lru_cache#functools.lru_cache) in the standard library?

async-aggregators/src/utils/apply_color_swap.py
```diff
@@ -0,0 +1,49 @@
...
+    if not (_img := _image_cache.get(img_path)):
+        _img = Image.open(img_path)
+        _image_cache.put(img_path, _img)
+    img = _img
```
camcaswell, line 33:
```suggestion
    if not (img := _image_cache.get(img_path)):
        img = Image.open(img_path)
        _image_cache.put(img_path, img)
```

async-aggregators/lib/pixelate_and_swap/pixelate_and_swap.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Use the masks to swap the colors. For pixels that match the mask1 (originally color1),
+    # set their value to color2.
+    img_array[mask1] = color2
+
+    # Similarly, for pixels that match mask2 (originally color2), set their value to color1.
+    img_array[mask2] = color1
+
+    # Convert the numpy array with swapped colors back to a PIL Image and return it.
+    return Image.fromarray(img_array)
```
camcaswell, line 76: This file is a little over-commented. Explaining what the K-means clustering is doing in the other function is helpful, but using 3 lines to explain a mask is excessive.

async-aggregators/src/main.py
```diff
@@ -0,0 +1,40 @@
...
+    if len(args) == 0:
+        level = 1
+    elif len(args) == 1:
+        try:
+            level = int(args[0])
+        except ValueError:
+            raise ValueError("First argument is not an integer")
+    else:
+        raise ValueError("Too many arguments specified")
+    if level < 1 or level > max_level:
```
camcaswell, line 27:
```suggestion
    if not 1 <= level <= max_level:
```

async-aggregators/src/main.py
```diff
@@ -0,0 +1,40 @@
...
+        raise ValueError(
+            f"Invalid level number. Please pick a number between 1 and {max_level}"
+        )
```
camcaswell, line 30: Your tox file specifies 119 as the col limit, it's good for everyone to be on the same page for formatting style decisions.
```suggestion
        raise ValueError(f"Invalid level number. Please pick a number between 1 and {max_level}")
```

async-aggregators/src/control_panel.py
```diff
@@ -0,0 +1,225 @@
...
+)
+
+from src.utils.q_custom_slider import QCustomSlider
+
+
+class NoDragSlider(QSlider):
+    """QSlider subclass to disable dragging."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super(NoDragSlider, self).__init__(*args, **kwargs)
```
camcaswell, line 17: This is the old way of calling `super`.
```suggestion
        super().__init__(*args, **kwargs)
```

async-aggregators/src/window.py
```diff
@@ -0,0 +1,127 @@
...
+        self.level = level
+        self.screen_size = QApplication.primaryScreen().size()
+        self._init_ui()
+
+    def _init_ui(self) -> None:
+        """Initialize the User Interface"""
+        self.window_name = "Digital Shadows - Async Aggregators"
+        self.setWindowTitle(f"{self.window_name} - Level {self.level.level_number}")
+
+        icons_dir_path = Path(Path(__file__).parent, "icons")
```
camcaswell, line 28:
```suggestion
        icons_dir_path = Path(__file__).parent / "icons"
```

async-aggregators/src/window.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return frame
+
+    def _create_tabbed_controls(self) -> QStackedLayout:
+        """Create the tabbed controls layout"""
+        layout = QStackedLayout()
+
+        for filter_item in self.level.filters:
+            if filter_item[1] is not None:
+                layout.addWidget(filter_item[1])
```
camcaswell, line 96: It's hard to tell what this is doing. Refactoring the `FilterList` as mentioned in another comment would help, but also a short comment and/or assigning a name would go a long way:
```py
for _,control_panel,_ in self.level.filters:
```

async-aggregators/src/window.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return layout
+
+    def _update_secret_code(self, input_code: str) -> None:
```
camcaswell, line 100: This name is a bit misleading since what this method actually does is go to the next level.

async-aggregators/src/window.py
```diff
@@ -0,0 +1,127 @@
...
+                return
+            else:
+                # If the input matches the answer, show a notification
+                msg_box = QMessageBox(self)
+                msg_box.setWindowTitle("Success")
+                msg_box.setText("\nCorrect secret code.\n")
+                msg_box.setIcon(QMessageBox.Icon.Information)
+                msg_box.exec()
+
+                self.level.level_up()
```
camcaswell, line 126: RE the other comment about the `Level` class:

`level_up` should be a method of whatever is tracking the gamestate and/or the main app window. It should create/load a *new* `Level`, and create/load whatever GUI elements that entails, including signals and callback functions connected to specific input elements. The `Level` object should not be mutating itself to represent a new level.

async-aggregators/src/filter.py
```diff
@@ -0,0 +1,203 @@
...
+                    "vertical wave",
+                    "horizontal spike",
+                    "vertical spike",
+                    "explode",
+                    "image_label_w",
+                    "image_label_h"
+            ):
+                args_for_filter[key] = value
+        return apply_motion(args_for_filter)
+    return QPixmap()
```
camcaswell, line 203: On seeing this my first question is "Why is something calling this one dispatch function instead of just calling the right function?" After chasing it up the call-stack a bit, I think this is the result of design decisions made elsewhere so for now I'll just keep it general to avoid trying to outline a whole refactor.

Input elements should emit a signal to their control panel. The control panel should call the appropriate image-transformation callback and pass it the current state of the GUI (the values for the various sliders, buttons, etc). Each control panel should be constructed for a specific level, so the appropriate callback functions are "baked into" the control panel.

The control panel knows which callback it uses, so this function dispatch is unnecessary. The callback functions know what state information to expect, so the `args` dict is unnecessary (see other comment about this pattern in general).

In other words we're replacing parsing strings dynamically when a button is hit with explicitly defining the logic for the GUI elements when they're created.

async-aggregators/src/control_panel.py
```diff
@@ -0,0 +1,225 @@
...
+            self.combo_boxes.append((combo_box1, combo_box2))
+        # Adding combo box buttons
+        for button_text in widget_info.get("combo_box_buttons", []):
+            button = QPushButton()
+            button.setText(button_text)
+            # Use default arguments in the lambda to capture the current values of combo_box1 and combo_box2
+            # Assuming one button per pair of dropdowns.
+            cb1, cb2 = self.combo_boxes.pop(
+                0
+            )  # Pop the first pair of QComboBoxes from the list
```
camcaswell, line 109: Don't let your formatter be your enemy. (Also this comment could really just be deleted since it doesn't really add anything.)
```suggestion
            # Pop the first pair of QComboBoxes from the list
            cb1, cb2 = self.combo_boxes.pop(0)
```

async-aggregators/src/utils/apply_unmask_reverse_ishihara.py
```diff
@@ -0,0 +1,57 @@
...
+        unmasked_image_rgb = cv2.cvtColor(unmasked_image, cv2.COLOR_BGR2RGB)
+
+        # Convert the numpy array (OpenCV image) to a QImage
+        height, width, channel = unmasked_image_rgb.shape
+        bytes_per_line = 3 * width
+        q_image = QImage(unmasked_image_rgb.data, width, height, bytes_per_line, QImage.Format.Format_RGB888)
+
+        # Convert the QImage to QPixmap
+        pixmap = QPixmap.fromImage(q_image)
+        resized_pixmap = pixmap.scaled(args["image_label_w"], args["image_label_h"])
```
camcaswell, line 52: This is a good candidate to be abstracted into a function.

async-aggregators/src/utils/q_custom_slider.py
```diff
@@ -0,0 +1,133 @@
...
+)
+
+
+class QCustomSlider(QWidget):
+    """A custom slider with tick labels"""
+
+    def __init__(self, minimum: int, maximum: int, interval: int = 1,
+                 orientation: Qt.Orientation = Qt.Orientation.Horizontal,
+                 labels: list | None = None, p0: typing.Any = 0, parent: typing.Any = None,
+                 suppress_mouse_move: bool = False) -> None:
```
camcaswell, line 16: Personally I think this is easier to read, but opinions vary.
```suggestion
    def __init__(self,
        minimum: int, maximum: int,
        interval: int = 1,
        orientation: Qt.Orientation = Qt.Orientation.Horizontal,
        labels: list | None = None,
        p0: typing.Any = 0,
        parent: typing.Any = None,
        suppress_mouse_move: bool = False,
    ) -> None:
```

async-aggregators/src/utils/q_custom_slider.py
```diff
@@ -0,0 +1,133 @@
...
+        levels = range(minimum, maximum + interval, interval)
+
+        if labels is not None:
+            if not isinstance(labels, (tuple, list)):
+                raise Exception("<labels> is a list or tuple.")
+            if len(labels) != len(levels):
+                raise Exception("Size of <labels> doesn't match levels.")
+            self.levels = list(zip(levels, labels))
+        else:
+            self.levels = list(zip(levels, map(str, levels)))
```
camcaswell, line 28: This version is more lenient on what types it will accept for `labels`, which could be good or bad. In general though, it's best to embrace the duck-typing and let the user pass whatever they want as long as it works. This version also has fewer branches, making it easier to read.
```suggestion
        if labels is None:
            labels = map(str, levels)
        labels = list(labels)
        if len(labels) != len(levels):
            raise ValueError("Size of <labels> doesn't match levels.")
        try:
            self.levels = list(zip(levels, labels))
        except TypeError:
            raise TypeError("<labels> is not iterable.")
```

async-aggregators/src/utils/q_custom_slider.py
```diff
@@ -0,0 +1,133 @@
...
+        style = self.sl.style()
+        painter = QPainter(self)
+        st_slider = QStyleOptionSlider()
+        st_slider.initFrom(self.sl)
+        st_slider.orientation = self.sl.orientation()
+
+        length = style.pixelMetric(QStyle.PixelMetric.PM_SliderLength, st_slider, self.sl)
+        available = style.pixelMetric(QStyle.PixelMetric.PM_SliderSpaceAvailable, st_slider, self.sl)
+
+        for v, v_str in self.levels:
```
camcaswell, line 84: Should probably be more descriptive names, e.g. `slider_value`, `slider_label`.

async-aggregators/lib/reverse_ishihara/reverse_ishihara.py
```diff
@@ -0,0 +1,106 @@
...
+    if message:
+        font_path = "MontserratRegular-BWBEl.ttf"  # Point to a TTF or OTF font file on your machine
+        font = ImageFont.truetype(font_path, int(image.height * 0.5))
+
+        left, top, right, bottom = font.getbbox(message)
+        text_width, text_height = right - left, bottom - top
+
+        text_img = Image.new('1', (text_width, text_height), color='white')
+        text_draw = ImageDraw.Draw(text_img)
+        text_draw.text((-left, -top), message, fill='black', font=font)  # Notice the negative offsets
```
camcaswell, line 39: These comments sound like they're for a tutorial. Make sure to take your expected audience into account when writing your code and comments.

camcaswell (commented):
Ahoy Async Aggregators, congrats on making the top 10! I'm glad I pulled your project to review because I particularly enjoyed it whlie judging.

I only made comments where I had something to bring up, so don't think I didn't like anything about your code. I actually thought the code quality was pretty high overall.

## README
Installation was very smooth (makes a good first impression when judging a bunch of projects ;p), and the commit history GIF was a nice touch.

## Teamwork
It looks like everyone contributed. 62 PRs is quite a lot! Commit messages were mostly good; a few were a little subpar but that's not unusual. If anyone wants a refresher check out [How to Write a Git Commit Message](https://cbea.ms/git-commit/).

## Code Style
Some of the things I commented on are nitpicky or simply matters of opinion. In those cases I'm just alerting you to alternative options. Overall, like I said, the code quality was pretty high. There were a couple places that needed more descriptive variable names, and there were some whitespace choices that looked like they might have been made by an overzealous autoformatter.

I think the biggest issue that stood out to me was the use of dictionaries instead of keyword arguments in some places. Passing arguments like that makes it hard to trace how data is moving through your program and what arguments functions expect.

## Code Organization
The big problem here was the `Level` class and (what I'm guessing are) the downstream design decisions. I made comments in several places regarding those, but let me try to summarize and unify them here.

The `Level` class is a singleton that represents the gamestate rather than a particular level. As such it has to keep track of the data for each level and mutate itself when the user goes on to the next level.

Similarly, the GUI does not have a strong separation of logic for the different levels. I can't be too specific here because I'm not super familiar with pyqt6, but ideally the level would provide the GUI with callbacks to interpret the current state of the puzzle. Instead there is one `apply_filter` function that handles this logic by parsing strings. I'm guessing this led to the usage of opaque "args" dicts for the filter functions.

There was some unused code scattered around, and in a real project I would expect some logging, but these things can be chalked up to the time pressure in a code jam.

## Summary
I thought you did a great job, both when I was sweating on some of the puzzles and when I was reading your code.

If you have any questions or *objections* feel free to leave comments and I'll try to reply in a reasonably timely manner.

camcaswell (commented):
Ahoy Async Aggregators, congrats on making the top 10! I'm glad I pulled your project to review because I particularly enjoyed it whlie judging.

I only made comments where I had something to bring up, so don't think I didn't like anything about your code. I actually thought the code quality was pretty high overall.

## README
Installation was very smooth (makes a good first impression when judging a bunch of projects ;p), and the commit history GIF was a nice touch.

## Teamwork
It looks like everyone contributed. 62 PRs is quite a lot! Commit messages were mostly good; a few were a little subpar but that's not unusual. If anyone wants a refresher check out [How to Write a Git Commit Message](https://cbea.ms/git-commit/).

## Code Style
Some of the things I commented on are nitpicky or simply matters of opinion. In those cases I'm just alerting you to alternative options. Overall, like I said, the code quality was pretty high. There were a couple places that needed more descriptive variable names, and there were some whitespace choices that looked like they might have been made by an overzealous autoformatter.

I think the biggest issue that stood out to me was the use of dictionaries instead of keyword arguments in some places. Passing arguments like that makes it hard to trace how data is moving through your program and what arguments functions expect.

## Code Organization
The big problem here was the `Level` class and (what I'm guessing are) the downstream design decisions. I made comments in several places regarding those, but let me try to summarize and unify them here.

The `Level` class is a singleton that represents the gamestate rather than a particular level. As such it has to keep track of the data for each level and mutate itself when the user goes on to the next level.

Similarly, the GUI does not have a strong separation of logic for the different levels. I can't be too specific here because I'm not super familiar with pyqt6, but ideally the level would provide the GUI with callbacks to interpret the current state of the puzzle. Instead there is one `apply_filter` function that handles this logic by parsing strings. I'm guessing this led to the usage of opaque "args" dicts for the filter functions.

There was some unused code scattered around, and in a real project I would expect some logging, but these things can be chalked up to the time pressure in a code jam.

## Summary
I thought you did a great job, both when I was sweating on some of the puzzles and when I was reading your code.

If you have any questions or *objections* feel free to leave comments and I'll try to reply in a reasonably timely manner.

PR #2, opened 2023-09-28
Dynamic Typists

janine9vn:


dynamic-typists/app/README.md
janine9vn, line 1: This type of documentation is very impressive and thorough. Having READMEs in the differnt subfolders to dive deeper into the details of each is very much appreciated.

dynamic-typists/app/picture.py
```diff
@@ -0,0 +1,91 @@
...
+        y = img_height // 2 - font_height // 2
+
+        # create watermark picture
+        watermark = Image.new(mode="RGB", size=self.image.size)
+        watermark_draw = ImageDraw.Draw(watermark)
+        watermark_draw.text((x, y), text, fill=(150, 150, 150), font=font)
+        watermark = watermark.rotate(45)
+
+        # chose contrast based on the originals image brightness
+        watermark_contrast = int(self._get_brightness() * 0.8)
```
janine9vn, line 75: This is nicely done and a helpful comment as well.

dynamic-typists/app/scrambler.py
```diff
@@ -0,0 +1,145 @@
+"""Scramble functions for scrambling images in different ways."""
```
janine9vn, line 1: I don't think a comment like this at the top of files is necessary. The name of this file and the functions contained within are enough imo.

dynamic-typists/app/scrambler.py
```diff
@@ -0,0 +1,145 @@
...
+
+def scramble_grid(picture: Picture) -> None:
+    """Split an Image up into tiles, rearrange them and rotate each tile randomly."""
+    num_of_tiles = 4
+    co_ordinates = [
+        (0, 0, 512, 512),
+        (512, 0, 1024, 512),
+        (0, 512, 512, 1024),
+        (512, 512, 1024, 1024),
+    ]
```
janine9vn, line 32: This seems fairly hardcoded for specific image sizes. Is the picture given guaranteed to be provided in the correct size?

dynamic-typists/app/server.py
```diff
@@ -0,0 +1,209 @@
...
+
+    scrambler = random.choice(SCRAMBLERS)
+
+    match scrambler:
+        case "rows":
+            scramble_rows(picture)
+        case "grid":
+            scramble_grid(picture)
+        case "circle":
+            scramble_circle(picture)
```
janine9vn, line 121: Not the most ideal usage of match-case, but I like the readability it provides here even if it's equivalent to if-elif statements.

dynamic-typists/app/server.py
```diff
@@ -0,0 +1,209 @@
...
+
+    # Create a session and store it.
+    session_id = str(uuid4())
+
+    SESSIONS[session_id] = SessionData(
+        picture=picture,
+        scrambler=scrambler,
+        session_id=session_id,
+        tiles_b64=tiles_b64,
+    )
```
janine9vn, line 134: This is a nice usage of sessions and also pydantic for the SessionData.

dynamic-typists/frontend/controllers/app.py
```diff
@@ -0,0 +1,214 @@
...
+
+        This method fetches a new CAPTCHA challenge from the server, which comprises a type
+        (e.g., 'grid', 'rows', 'circles') and a set of associated tile images. Depending on
+        the CAPTCHA type, it transitions to the corresponding controller and instructs it
+        to render the fetched tiles.
+        """
+        captcha = await fetch_tiles()
+
+        controller_name = captcha["type"]
+        tiles = captcha["tiles"]
```
janine9vn, line 166: How certain is it that you'll always have those as part of the `catcha` response?

dynamic-typists/frontend/controllers/app.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.code_input.value = ""
+        self.set_controller(controller_name)
+
+        if self.active_controller is not None:
```
janine9vn, line 171: will `active_controller` ever be `None`? It loads it a few lines up, taking the value from `fetch_tiles()` which tracing it back to `server.py` seems to always return a proper value.

dynamic-typists/frontend/controllers/transform_controller.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, element: JsDomElement) -> None:
+        """Create a new `TransformController` with the provided DOM element.
+
+        Parameters
+        ----------
+        element:
+            The DOM element to which the transformations will be applied.
+        """
+        self.element: JsDomElement = element
+        self.transforms: dict[Transform, str] = {}
```
janine9vn, line 35: This feels a little bit misleading =P
From what I can tell, there are instances where `self.transforms` remains an empty dict, or at least it is unclear when/where/how `self.transforms` becomes populated with at least 1 `Transform`.

dynamic-typists/frontend/controllers/app.py
```diff
@@ -0,0 +1,214 @@
...
+
+import json
+from functools import partial
+from typing import TYPE_CHECKING, Protocol
+
+from pyodide.http import pyfetch
+
+from .drag_drop_grid_controller import DragDropGridController
+from .image_grid_controller import ImageGridController
+from .rotating_images_controller import RotatingImagesController
```
janine9vn, line 11: I would prefer these (and others in the other files) to be absolute imports, especially since you use absolute imports below. In general, absolutely imports are preferred and I don't think the few extra names to add here would be too much.

dynamic-typists/frontend/controllers/app.py
```diff
@@ -0,0 +1,214 @@
...
+
+    response = await pyfetch(
+        "api/solution",
+        method="POST",
+        headers={"Content-Type": "application/json"},
+        body=json.dumps(body),
+    )
+
+    if not response.ok:
+        # Need to properly handle errors
```
janine9vn, line 127: Unsure what the purpose of this comment is here. Is it a reminder to do something later? Is it a note to future developers?

dynamic-typists/frontend/controllers/click_rotation_controller.py
```diff
@@ -0,0 +1,45 @@
...
+    def _on_left_click(self, event: MouseEvent) -> None:
+        """Handle the left-click event to rotate the element.
+
+        Parameters
+        ----------
+        event:
+            The mouse event associated with the click action.
+        """
+        event.preventDefault()
+        self.step_clockwise()
```
janine9vn, line 45: Is the only option stepping clockwise? Itt would be nice is an alternative click or key-board combo could rotate this counter-clockwise. The docstrings in this function also do not note anywhere the clockwise-only movement when you left click.

dynamic-typists/frontend/controllers/drag_drop_grid_controller.py
```diff
@@ -0,0 +1,276 @@
...
+
+            # If source is right before target
+            if source_next_sibling == target:
+                self.root.insertBefore(target, source)
+            # If target is right before source
+            elif target_next_sibling == source:
+                self.root.insertBefore(source, target)
+            else:
+                insert_or_append(source, target_next_sibling)
+                insert_or_append(target, source_next_sibling)
```
janine9vn, line 269: I don't think `handle_insert` and `handle_swap` need to be functions here. Instead, have the logic in an if/else statement depending if you're swapping or inserting. `insert_or_append` should probably be a separate non-nested method as well (I generally dislike nested functions if the primary purpose is encapsulation)

This suggestion is courtesy of Scofflaw:
```diff
         if source_index == target_index:
             return

-        # Helper function to insert or append
-        def insert_or_append(element: JsDomElement, reference: JsDomElement | None) -> None:
-            if reference is not None:
-                self.root.insertBefore(element, reference)
-            else:
-                self.root.appendChild(element)
-
-        def handle_insert() -> None:
+        if self.drop_behavior == "insert":
             # If dragging from left to right
             if source_index < target_index:
-                insert_or_append(source, target.nextSibling)
+                self.insert_or_append(source, target.nextSibling)
             # If dragging from right to left
             else:
                 self.root.insertBefore(source, target)

-        def handle_swap() -> None:
+        else:
             source_next_sibling = source.nextSibling
             target_next_sibling = target.nextSibling

            # If source is right before target
            if source_next_sibling == target:
                self.root.insertBefore(target, source)
            # If target is right before source
             elif target_next_sibling == source:
                 self.root.insertBefore(source, target)
             else:
-                insert_or_append(source, target_next_sibling)
-                insert_or_append(target, source_next_sibling)
+                self.insert_or_append(source, target_next_sibling)
+                self.insert_or_append(target, source_next_sibling)

-        drop_behaviors: dict[DropBehavior, Callable[[], None]] = {
-            "insert": handle_insert,
-            "swap": handle_swap,
-        }

-        drop_behaviors[self.drop_behavior]()
+    # Helper method to insert or append
+    def insert_or_append(self, element: JsDomElement, reference: JsDomElement | None) -> None:
+        if reference is not None:
+            self.root.insertBefore(element, reference)
+        else:
+            self.root.appendChild(element)
```

dynamic-typists/app/server.py
```diff
@@ -0,0 +1,209 @@
...
+
+def random_image() -> Path:
+    """Choose and return a random image path from the resource directory.
+
+    Returns
+    -------
+    Path
+        The path to a randomly selected image.
+    """
+    images = list(itertools.chain(*(list(RESOURCES.glob(glob)) for glob in GLOBS)))
```
janine9vn, line 62: This is doing quite a bit. I think this can be simplified to the following:
```py
images = []
for glob in GLOBS:
    images.extend(RESOURCES.glob(glob))
```
It's not a one-liner, but I think makes it easier to reason about and involves fewer calls/operations.

dynamic-typists/app/scrambler.py
```diff
@@ -0,0 +1,145 @@
...
+        tile_img = tile_img.rotate(tile_rotations[tile])
+        picture.tiles[tile] = Tile(
+            tile_img,
+            (co_ordinates[tile][0], co_ordinates[tile][1]),
+            tile_rotations[tile],
+        )
+
+
+def scramble_circle(picture: Picture) -> None:
+    """Split an Image up into circles and rearrange them."""
```
janine9vn, line 69: This is doing a lot more than this docstring implies. I would expand on this to explain a bit more about what the function will do, especially the masking aspects.

You also have a bit of repeated code, so while not immediately a problem may be indicative that some of the code can be turned into helper functions.

dynamic-typists/app/scrambler.py
```diff
@@ -0,0 +1,145 @@
...
+            tile_order[0], tile_order[1] = tile_order[1], tile_order[0]
+
+        if current_choice == "left":
+            tile_order[0], tile_order[2] = tile_order[2], tile_order[0]
+
+        if current_choice == "bottom":
+            tile_order[2], tile_order[3] = tile_order[3], tile_order[2]
+
+        if current_choice == "right":
+            tile_order[1], tile_order[3] = tile_order[3], tile_order[1]
```
janine9vn, line 52: There is a non-trivial chance that the shuffled order will be the same as the original order. I would add a check at somepoint to make sure you actually did shuffle the input tiles.

Also, this is currently hard-coded to be broken up into 4 tiles. I think a good future improvement would to make this be scalable, to accomodate any square number of tiles.

dynamic-typists/app/server.py
```diff
@@ -0,0 +1,209 @@
...
+    session_id = str(uuid4())
+
+    SESSIONS[session_id] = SessionData(
+        picture=picture,
+        scrambler=scrambler,
+        session_id=session_id,
+        tiles_b64=tiles_b64,
+    )
+
+    response.set_cookie(key="session_id", value=str(session_id))
```
janine9vn, line 136: `session_id` is already a string (line 127), so does not need to be converted again here.

dynamic-typists/app/solver.py
```diff
@@ -0,0 +1,72 @@
...
+        An instance of the Picture class representing the scrambled CAPTCHA.
+
+    Returns
+    -------
+    Solution
+        A list with the rotation needed to put the ring in the right spot.
+    """
+    solution = []
+    num_tiles = len(picture.tiles)
+    for ring_position in range(1, num_tiles):
```
janine9vn, line 69: I think this can be condensed, having `num_tiles` as its own variable isn't doing much here.

dynamic-typists/frontend/controllers/rotating_images_controller.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        self.root.classList.add("rotating-images")
+        background, *rotatable = rotatable
+
+        background_element = document.createElement("img")
+        background_element.src = f"data:image/png;base64,{background}"
+        background_element.classList.add("background-image")
+        self.root.appendChild(background_element)
+
+        for image in reversed(rotatable):
```
janine9vn, line 60: Why reversed here? It's not clear why this needs to be reversed.

dynamic-typists/frontend/controllers/rotating_images_controller.py
```diff
@@ -0,0 +1,110 @@
...
+    @property
+    def solution(self) -> list[float]:
+        """Get the current solution as a list of the degrees by which each element is rotated.
+
+        Returns
+        -------
+        List[float]
+            List of rotation values in degrees.
+        """
+        return [controller.current_rotation for controller in reversed(self._controllers)]
```
janine9vn, line 110: It seems the order of the list of rotation values is important, but it's not clear which direction it should be or which direction it's starting in.

dynamic-typists/frontend/main.py
```diff
@@ -0,0 +1,126 @@
...
+        if solved:
+            confirm_button.classList.add("solved")
+            confirm_button_text.innerText = "SOLVED"
+            remove_event_listener(confirm_button, "click", handle_post_solution)  # type: ignore
+        else:
+            confirm_button_text.innerText = "RETRY"
+
+    add_event_listener(confirm_button, "click", handle_post_solution)  # type: ignore
+    add_event_listener(refresh_button, "click", handle_load_captcha)  # type: ignore
+    add_event_listener(restart_button, "click", lambda _: app.reset())
```
janine9vn, line 123: Rather than have a lambda to "eat" the argument, I would rather the `reset()` function itself in `app.py` discard/ignore all arguments given to it. `reset()` is only called here, so it feels a bit odd to know it will be passed an unnecessary event and that you have to go out of your way to discard it with a lambda.

dynamic-typists/frontend/main.py
```diff
@@ -0,0 +1,126 @@
...
+        ----------
+        _ : Event
+            The event object, which is not used in this function but is typically passed by
+            event handlers.
+
+        Notes
+        -----
+        The function utilizes a nonlocal variable `is_loading_captcha` to prevent multiple
+        concurrent requests when the user repeatedly clicks the button. This ensures that the
+        application does not request another captcha until the previous one has been loaded.
```
janine9vn, line 50: Smart choice!

janine9vn (commented):
Hello! Thanks for your patience as we get this review sorted out in GH format.

## Documentation & README
The documentation throughout this project is of a very high level and impressive. This project would have been much more difficult to review with out. The READMEs in each section are helpful for explaining the layout and some decisions made. In some cases it may even be *too* thorough and in-depth, although I don't think you'll find anyone complaining. There are still a handful of instances that could use a bit more explanation or comments, but by and far the documentation for this is very well done.

The module docstrings can either be removed or be expanded on. Right now they're sitting in a bit of an uncertain zone of having some basic information, but not enough to prevent someone from going elsewhere to get a deeper-than-surface-level understanding of what the module is doing.

Also shoutout for the inclusion of the `js.pyi`, it was helpful for this review and to stop my IDE from complaining.

## Code Quality & Organization
The code is also very well done. There are a few instances where I think some things are unnecessarily complicated, a little too long for a single function, or there are nested functions that can be broken apart / simplified. I'm not familiar with pyodide in the slightest, but the this code was still (mostly) easy to reason about.

That being said, I did get a bit lost in the sauce trying to make sense and understand the different controllers. That was one area where I do wish there a bit more documentation explaining what controller is used where and why. For instance, the `drag_drop_grid_controller` and the `image_grid_controller` sound like they do the exact same thing but worded slightly differently. Turns out that `image_grid_controller` does user `drag_drop_grid_controller`, but it took a little bit to trace that back and keep track of it all. So either a combination of a slight re-organization and slight-renaming or a document/diagram illustrating how they're all used could go a long way.

## Commits
21 commits is not a lot of commits for a project this size. That number alone, looking at the scale and scope of this project, implies that some commits are likely bigger than they should be. Digging into it a bit, the true number of commits is lost at the Pull Request stage. The individual commit information is retained in the pull request, but when I'm viewing it locally there are just these giant commits that do a lot that I can't break apart into more manageable peices. So I recommend following an atomic git commit approach in the future. I like recommending [Make Atomic Git Commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/) for further reading about atomic git commits and why they are beneficial.

## Summary
Overall, this project is very well done. It has really solid code quality, excellent documentation, and the actual execution of it is impressive. Good job y'all~

dynamic-typists/frontend/controllers/rotating_images_controller.py
thijs-franck: This definitely requires some explanation. 

We receive images from the API in the order: `[background_image, smallest_ring, ..., largest_ring]`. In HTML rendering, the last element in the list appears on top. Since each ring must remain selectable for user rotation, and considering the transparent part of a ring is still part of its clickable element, we need to ensure that smaller rings aren't obscured by larger ones. Thus, by reversing the list, we ensure smaller rings are rendered last (and therefore on top), keeping them selectable.

dynamic-typists/frontend/controllers/rotating_images_controller.py
thijs-franck: Because we create the controllers in reverse order, we need to again reverse the order here to build the solution that the API expects.

dynamic-typists/app/server.py
ooliver1: Preferred match for this since the performance impact is negligible

PR #3, opened 2023-09-28
Expected Indents

janine9vn:


expected-indents/src/static/script_game.js
```diff
@@ -0,0 +1,152 @@
...
+			.then(response => response.json())
+			.then(data => {
+				/*
+				I forgot exactly how we decided the API response should be... so I assume it's like this:
+				{
+					"word": "w_r_",
+					"lives": 5,
+					"image": "https://example.com/image.png"
+				}
+				*/
```
etrotta, line 35: This comment is a bit worrying, both it existing in first place as well as it still being there in the PR

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        length (int, default=6): The length of the random id. Should be 1 or more.
+
+    Raises:
+        AssertionError: Length can't be 0 or lower.
+
+    Returns:
+        str: A random uid.
+    """
+    characters = string.ascii_letters + string.digits
+    chars = [secrets.choice(characters) for _ in range(length)]
```
etrotta, line 24: Why `secrets.choice` instead of `random.choices`?

Personally I'd probably just use an `uuid.UUID4` instead to be honest

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+    return random_string
+
+
+class Game:
+    """game class"""
+
+    def __init__(self, lives=6):
+        self.word = self._get_random_word()
+        # print(self.word)
+        self.total_lives = lives  # const
```
etrotta, line 35: Unused variable? (`total_lives`)

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        # but this is just a constant to calculate the reward
+
+    def _get_random_word(self):
+        """Generate a random word from the wordlist
+
+        Returns:
+            a random word from nouns-clear.txt
+        """
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        file_path = os.path.join(script_dir, 'nouns-clear.txt')
```
etrotta, line 62: Use `pathlib` over `os.path`.
Arguably personal preferences, specially since it should work in any OS using either, but I still recommend `pathlib` over it

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        with open(file_path) as nounfile:
```
etrotta, line 63: Preferably specify the mode and encoding whenever you're opening a file

The encoding doesn't matters much when all of the text in the file is ASCII, but the default encoding is currently platform dependent, the default encoding for Windows may change in the future, and I personally believe that it's better to explicit state the mode even when using the default ('r')

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        str: A random uid.
+    """
+    characters = string.ascii_letters + string.digits
+    chars = [secrets.choice(characters) for _ in range(length)]
+    random_string = ''.join(chars)
+    return random_string
+
+
+class Game:
+    """game class"""
```
etrotta, line 30: May as well not have a docstring over just the name of the class. The comments in `init` could be a bit tidier, but are passable

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns:
+            a random word from nouns-clear.txt
+        """
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        file_path = os.path.join(script_dir, 'nouns-clear.txt')
+        with open(file_path) as nounfile:
+            rword = random.choice(nounfile.readlines()).strip()
+        return rword
+
+    def _construct_word_with_underscores(self):
```
etrotta, line 67: nothing wrong with it, but just a side note: some might prefer to do that in one line, but nothing wrong with doing it explicitly with a normal loop, maybe even better this way.

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+        The reward is 0 when no letters are guessed
+        The reward is self.m when half of the letters are guessed
+        The reward is self.u when all letters are guessed
+
+        """
+        x = sum([letter in self.guessed_letters for letter in self.word])/len(self.word)
+        k = (self.u-self.m)/self.m
+        coeff = self.m / (k-1)
+        self.reward = int(coeff * (k**(2*x) - 1)) + 1
+
```
etrotta, line 95: use more descriptive variable names. If the formula you are using has a name, document it

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+    def _calculate_punishment(self):
+        """Calculate the punishment"""
+        subgrid_size = self.size // self.grid
+
+        for i in range(self.grid):
+            for j in range(self.grid):
+                if np.random.rand() < self.p:
```
etrotta, line 102: create a rng instance like `rng = np.random.default_rng()` then use `rng.random()` instead of calling `np.random.rand` directly

See https://numpy.org/doc/stable/reference/random/index.html for "why"

expected-indents/src/game.py
```diff
@@ -0,0 +1,163 @@
...
+                    ] = 0
+
+    def play(self, guessed_letter):
+        """Play the game by guessing a letter.
+
+        Args:
+            guessed_letter (str): The letter guessed by the user. Should be a single letter.
+
+        Returns:
+            dict: {
```
etrotta, line 115: Consider using a TypedDict or a dataclass for these cases, when you have a dictionary with fixed keys, or even something like a Pydantic model

etrotta: Also why return all of `lives`, `delta_lives` and `feedback` instead of only feedback or just lives+feedback?

expected-indents/src/server.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    cgame = games.get(game_id)
+    if cgame is None:
+        return {'error': 'game does not exist'}
+    img_io = BytesIO()
+    np_image = cgame.get_image()
+    pil_img = Image.fromarray(np_image)
+    pil_img.save(img_io, 'JPEG', quality=70)
+    img_io.seek(0)
+    return send_file(img_io, mimetype='t')
```
etrotta, line 73: Why "t"? Unless it has some special meaning I do not know, you should use an actual mimetype or even just let flask infer it

expected-indents/src/server.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+app.run(debug=True, host='127.0.0.1', port=5000)
```
etrotta, line 76: Put the `app.run()` inside of an `if __name__ == "__main__"` block

Usually you should use a webserver like gunicorn or waitress instead of using flask's default debugging server if you were to deploy a flask program to production

expected-indents/src/svd.py
```diff
@@ -0,0 +1,90 @@
...
+        self.word = word
+        self._fetch_image(word)  # self.A is defined by this function
+        if len(self.A.shape) != 3:
+            raise ValueError("Image does not have 3 channels")
+        # separating the R, G and B channels to 2D matrices
+        self.R = self.A[:, :, 0]
+        self.G = self.A[:, :, 1]
+        self.B = self.A[:, :, 2]
+
+        if verbose:
```
etrotta, line 39: Use the `logging` module instead of determining whenever or not to print with an argument like `verbose` yourself
Logging manages that for you, by letting you set the logging level filter (error, warning, info, debug), either globally or per module. It is specially useful to have an unified API for that if you want to set the logging level for dependencies, but it's useful even for your own code once it gets somewhat big.

expected-indents/src/svd.py
```diff
@@ -0,0 +1,90 @@
...
+        S = np.diag(S)
+        self.BU, self.BS, self.BV = U, S, V
+
+    def _fetch_image(self, word):
+        """Fetch image from loremflickr API using word
+
+        Args:
+            word (str): Image search term
+        """
+        res = requests.get(f"https://loremflickr.com/512/512/{word}", timeout=10)
```
etrotta, line 63: Check the status of the response, e.g. at least call `raise_for_status()`, before trying to load it's content
Even if it works normally, you should always keep in mind that network requests could fail, for example the server could be overloaded and return a 500 error or not be accessible for whichever reason (failing to resolve the DNS record, you're offline, the server is offline, the server is literally on fire etc).

~~Also that `BytesIO` is redundant I think?~~ Never mind, seems like it is necessary
* the BytesIO in the line 64

expected-indents/src/svd.py
```diff
@@ -0,0 +1,90 @@
...
+        image = Image.open(BytesIO(res.content))
+        image = image.resize((self.size, self.size))
+        self.A = np.array(image)
+
+    def reduce(self, terms, type=np.uint8):
```
etrotta, line 68: avoid using built-in names such as `type` as a variable name, for example you could use `dtype` here matching numpy's terminology

expected-indents/src/svd.py
```diff
@@ -0,0 +1,90 @@
...
+        """Reduce the image to the specified number of terms
+
+        Args:
+            terms (int): Number of terms to keep
+            type (numpy type, optional): Type of the returned array. Defaults to np.uint8.
+
+        Returns:
+            _type_: _description_
```
etrotta, line 76: forgot to change this? (line 76)

expected-indents/src/server.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@app.route('/guess/<game_id>/<letter>')
+def make_move(game_id, letter):
+    """Process and make a move.
+
+    Args:
+        game_id (str): The id for the game in which the move was made.
+    Raises:
+        AssertionError: Game must exist, game_id must be a key the games dictionary.
```
etrotta, line 41: In multiple docstrings you have "Raises AssetionError", but you do not actually use `assert` in any of these places?
Side note: overall you might want to avoid using `assert` (not that you have any in this project rn) and normally raise exceptions instead, in part since `assert` is skipped if you use optimization command line arguments when running python (`-OO`), in part so that you and other people using your code can catch them based on their type with `except:`

expected-indents/src/server.py
```diff
@@ -0,0 +1,76 @@
+from io import BytesIO
+
+from flask import Flask, render_template, send_file
+from PIL import Image
+
+from game import Game, make_uid
+
+app = Flask(__name__)
+
+games = dict()
```
etrotta, line 10: Avoid using `global` variables, use [Flask's Application Context](https://flask.palletsprojects.com/en/2.3.x/appcontext/) instead.

Global variables are bug prone overall, specially in multi-threaded contexts
Also, you do not need to use the `global` keyword to _modify_ global variables, only to _overwrite_ them. Do not use global variables either way though.

etrotta:
Some notes that are not specific to any files, but overall recommendations:
(not requirements from the code jam, but personal recommendations from me)
- use unit tests
- use type hints on function definitions
- always use absolute imports over relative imports

Also, it might look like I reviewed extremely fast, but I took notes before officially starting the review - much of it was copy/pasting from the notes

expected-indents/README.md
```diff
@@ -0,0 +1,150 @@
...
+Here the word "answer" is the solution, and as you can see the player guesses 3 letters incorrectly, making random squares disappear from the image! However, when the player guesses correctly, the image gets slowly more comprehensible.
+
+## Instructions for use and Installation
+### From scratch:
+- Clone [the repo](https://github.com/BUSH222/cj-The-Expected-Indents)
+- Install the required libraries in `dev-requirements.txt`
+- Run `src/server.py` using python
+- In your browser, go to [`http://127.0.0.1:5000`](http://127.0.0.1:5000)
+### OR Visit the website:
+[http://picturepuzzlers.pythonanywhere.com/](http://picturepuzzlers.pythonanywhere.com/ "http://picturepuzzlers.pythonanywhere.com/")
```
etrotta, line 37: No need to use `[]()` as far as I know? Just put it normally

etrotta (commented):
forgot to actually Submit the review, sorry

PR #4, opened 2023-09-28
Itinerant Iterators

janine9vn:


itinerant-iterators/.github/workflows/tests.yaml
```diff
@@ -0,0 +1,40 @@
+# GitHub Action Workflow enforcing our code style.
```
ichard26, line 1: Make sure when you're copying and pasting any sort of code that you make the necessary modifications so it fits the context. It's too easy to end up with confusing comments like this :P

itinerant-iterators/.github/workflows/tests.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+name: Tests
+
+# Trigger the workflow on both push (to the main repository, on the main branch)
+# and pull requests (against the main repository, but from any repo, from any branch).
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
ichard26, line 11: I love to see workflow_dispatch to be used here. It's a great feature.

itinerant-iterators/LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2021 Python Discord
```
ichard26, line 1: We don't hold the copyright to your project's code, don't forget to change this next year if you're participating again :)

itinerant-iterators/tests/data/LICENSES
ichard26, line 1: I'm pleased to see the license information was maintained. You don't see that too often these days!

itinerant-iterators/samples/Pipfile
ichard26, line 1: You don't need this folder at all. :slightly_smiling_face:

itinerant-iterators/dev-requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+# Base tools
+flake8~=4.0.1
+isort~=5.10.1
+pre-commit~=2.17.0
+
+# Flake8 plugins, see https://github.com/python-discord/code-jam-template/tree/main#plugin-list
+flake8-docstrings~=1.6.0
+
+pytest
+black
```
ichard26, line 14: Pin the version of Black used to at least a major revision so you don't face sudden formatting changes when the new year rolls around. For more information, [see Black's stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy).

itinerant-iterators/justfile
ichard26, line 1: I love the use of a task runner. It makes getting up and running with a new project so much easier and faster.

Although, while just is a perfectly capable task runner, I would suggest looking into Tox, Nox, or even Hatch's own task runner. They have automatic environment provisioning (ie. setting up virtual environments and installing dependencies as necessary) that is tailored to Python's ecosystem. They may be more intuitive and provide a more streamlined experience compared to just.

ichard26 (commented):
Hello, I'm Richard and I'm the assigned reviewer for your project! Congratulations on submitting a top 10 project.

My review is structured in five parts:

1. Use of Git and GitHub + Project infrastructure (GHA and other developer tools)
2. Code review for the application
3. Code review for the server
4. Code review for the tests
5. Final comments (spoiler: it's going be pretty short, the juicy details will live in the four other parts :slightly_smiling_face:) 

**This review is the first part.** And before I get into it, remember, if you have any questions or want to discuss my feedback, please let me know!

## Use of Git and GitHub

The Git commit history is decent. I like how most commits are "atomic" and are reasonably sized and focused. However,  I would like to see more context on commits that have vaguer titles. For example, https://github.com/smileyface12349/itinerant-iterators/commit/ca0eb1c8d2e30909fbb3ba2a7e541e95e785dd2d 's commit message doesn't explain what was broken about the obfuscator... which makes it hard to tell what the commit does at a glance. 

Also, don't forget to squash or use interactive rebase to clean up local commits before pushing to avoid situations like this one:
image

I'm happy to see GitHub Issues used to keep track of work and discuss tasks. Keep that up! It's a skill you'll need a lot. If you want to take it further, I'd suggest using Pull Requests to encourage code review before a change is merged.

Further reading:

- https://cbea.ms/git-commit/
- https://about.gitlab.com/blog/2020/11/23/keep-git-history-clean-with-interactive-rebase/
- https://stackoverflow.com/questions/5308816/how-can-i-merge-multiple-commits-onto-another-branch-as-a-single-squashed-commit
- https://docs.github.com/en/get-started/quickstart/github-flow

## Project set up & infrastructure

The way the project is set up makes sense. Tooling like Black and pre-commit are appropriately used to maintain a consistent code style. Great job on using a GitHub Actions workflow to automatically run tests on each push.

I think the user experience would improve if the application was packaged into an installable Python package, but I'll discuss this further in Part 2.

itinerant-iterators/app/__main__.py
```diff
@@ -0,0 +1,4 @@
+from app.cli import main
```
ichard26, line 1: I don't have a preference between relative and absolute imports, but being consistent is important. This breaks running the application from a different directory that isn't the repository root. I'd suggest using relative imports as those seem to be a better fit (the project isn't installable):

```python
from .cli import main
```

itinerant-iterators/app/byteutils.py
```diff
@@ -0,0 +1,42 @@
...
+
+def iter_bits(bytes_: bytes) -> Generator[int, None, None]:
+    """Enumerate over each bit in the given bytes.
+
+    Args:
+        bytes_ (bytes): bytes to enumerate over
+
+    Yields:
+        Generator[int, None, None]: Return each bit as an int (0/1)
+    """
```
ichard26, line 12: Avoid repeating the argument and return type annotations in the docstring. An IDE should be able to use both to provide rich help. Down the road, it's very likely the docstring types will be out of date with the actual type annotations.

(This applies to the rest of the project, but this was the first example of it.)

itinerant-iterators/app/byteutils.py
```diff
@@ -0,0 +1,42 @@
...
+def set_bit(value: int, bit: int) -> int:
+    """Set the bit as position on the value.
+
+    Args:
+        value (int): Value to set bit on
+        bit (int): Bit position to set
+
+    Returns:
+        int: value with the bit at position `bit` set
+    """
```
ichard26, line 28: This is what I mean by unclear docstrings. It's good that they exist, but it's no good if they do little to help the reader with what the function does. How about this?

```suggestion
    """Set a bit in a byte to one.

    Args:
        value: Byte to set bit on
        bit: Bit index to set

    Returns:
        int: Value with the bit at :bit: location set. 
    """
```

itinerant-iterators/app/byteutils.py
ichard26, line 1: It's great that the bit manipulation functions are located in their own module, but this should probably live in `app.steganography` since the obfuscation code makes no use of this code.

itinerant-iterators/app/cli.py
ichard26, line 1: Excellent usage of Click. Command groups, parameter types and other Click features are effectively used to compose a clear CLI. The application's UI could still use some work though. Please see feedback written above.

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+@watermark.command("encode")
+@click.argument("text", type=str)
+@click.argument(
+    "infile",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.argument(
+    "outfile",
+    type=click.Path(exists=False, dir_okay=False, file_okay=True, writable=True),
+)
```
ichard26, line 46: It's great that Click's Path type is being used, but I have two suggestions:

- Don't forget to pass [`path_type=Path`](https://click.palletsprojects.com/en/8.1.x/api/#click.Path) so Click will convert the path into a `pathlib.Path` object instead of leaving it as a string (the parameter type annotations are technically incorrect due to this)
- Create `INPUT_FILE` and `OUTPUT_FILE` type constants that can be reused across all `@click.argument` and `@click.option` calls. This cuts down the repetition and makes the code easier to read. For example... 

```python
INPUT_FILE = click.Path(exists=True, file_okay=True, dir_okay=False, readable=True, path_type=Path)
OUTPUT_FILE = click.Path(exists=False, dir_okay=False, file_okay=True, writable=True, path_type=Path)

@watermark.command("encode")
@click.argument("text", type=str)
@click.argument("infile", type=INPUT_FILE)
@click.argument("outfile", type=OUTPUT_FILE)
def watermark_encode(text: str, infile: Path, outfile: Path):
    ...
```

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+    if result:
+        click.echo(result)
+    else:
+        click.echo(f"no encoded data found in {infile}")
+
+
+@cli.command("obfuscate")
+@click.option("--mode", "-m", type=click.Choice(["blur", "colour"]))
+@click.option("--colour", "-c", type=str, default="black", required=False)
+@click.option("--regex/--no-regex", type=bool, default=False)
```
ichard26, line 70: Click options can receive a `help` argument that will be emitted in help output. Use it to better document the various knobs the CLI has :)

```python
import click

@click.command
@click.option("--loud/--quiet", "yell", help="Shout (use uppercase) message.")
def main(yell: bool) -> None:
    click.echo("HELLO, WORLD!" if yell else "hello, world")

main()
```

```
❯ python test.py --help
Usage: test.py [OPTIONS]

Options:
  --loud / --quiet  Shout (use uppercase) message.
  --help            Show this message and exit.
```

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+@click.argument(
+    "infile",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.argument(
+    "outfile",
+    type=click.Path(exists=False, dir_okay=False, file_okay=True, writable=True),
+)
+@click.argument("filters", type=str, nargs=-1)
```
ichard26, line 79: `type=str` and `type=bool` are redundant. No need to specify them.

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+def obfuscate_cmd(
+    mode: str, colour: str, regex: bool, infile: Path, outfile: Path, filters: list[str]
+):
+    """Entrypoint to obfuscate command"""
+    # setup obfuscator
+    obfuscator = obfuscate.ColourBox(colour)
+    if mode == "blur":
+        obfuscator = obfuscate.BlurBox()
+    elif mode == "colour":
+        obfuscator = obfuscate.ColourBox(colour)
```
ichard26, line 89: Do not set a default obfuscator if you're going to overwrite it no matter what. If needed, raise an AssertionError as a fallback else clause.

```suggestion
    if mode == "blur":
        obfuscator = obfuscate.BlurBox()
    elif mode == "colour":
        obfuscator = obfuscate.ColourBox(colour)
    else:
        assert False, "unimplemented obfuscation mode"
```

itinerant-iterators/app/obfuscate/blur_box.py
```diff
@@ -0,0 +1,38 @@
...
+        self,
+        iterations: int = 20,
+        image_filter: Filter = ImageFilter.BoxBlur(radius=10),
+    ):
+        """Create a new blur obfuscator
+
+        Args:
+            iterations (int, optional): Number of blur iterations to apply to the bounding box. Defaults to 20.
+            image_filter (Filter, optional): Filter to apply for bluaring. Defaults to box blur blur with radius 10.
+        """
```
ichard26, line 21:
```suggestion
        Args:
            iterations (int, optional): Number of blur iterations to apply to the bounding box. Defaults to 20.
            image_filter (Filter, optional): Filter to apply for bluring. Defaults to box blur blur with radius 10.
        """
```
Perhaps using [codespell](https://github.com/codespell-project/codespell) (as needed or as a pre-commit hook) would be in order :slightly_smiling_face:

itinerant-iterators/app/obfuscate/blur_box.py
```diff
@@ -0,0 +1,38 @@
...
+        self.__iterations = iterations
+        self.__image_filter = image_filter
```
ichard26, line 23: Mangled "private" attributes are basically never used in Python, and they shouldn't be used here. They should be only used to avoid conflicts caused by subclassing, but since these attributes are unique to the class those kinds of conflicts can't happen anyway.

Use normal attributes instead, they're more readable too.

itinerant-iterators/app/obfuscate/colour_box.py
```diff
@@ -0,0 +1,40 @@
...
+    def hide(self, box: tuple[int, int, int, int], image: PILImage):
+        """Hide the pixels in the bounding box given in the image.
+
+        Args:
+            box (tuple[int, int, int, int]): Bounding box to obfuscate
+            image (PILImage): Image to obfuscate
+        """
+        for i in range(box[0], box[2]):
+            for j in range(box[1], box[3]):
+                image.putpixel((i, j), self.__colour)
```
ichard26, line 40: Use conventional x, y names for variables representing coordinates.
```suggestion
        for x in range(box[0], box[2]):
            for y in range(box[1], box[3]):
                image.putpixel((x, y), self.__colour)
```

itinerant-iterators/app/obfuscate/colour_box.py
```diff
@@ -0,0 +1,40 @@
...
+    """ColourBox is a form of obfuscation where it sets all pixels in a given bounding box to the given colour."""
+
+    def __init__(self, colour: tuple[int, int, int] | str):
+        """Create a new ColourBox obfuscator
+
+        Expects the bounding box to be valid for the image.
+
+        Args:
+            colour (tuple[int, int, int] | str): colour to set pixels in the bounding box to.
+            Can be a string e.g. "black", or an RGB tuple e.g. (255,255,255)
```
ichard26, line 17: Excellent argument documentation. This is exactly what I would like to see more.

itinerant-iterators/app/obfuscate/find_bounds.py
```diff
@@ -0,0 +1,39 @@
+import requests
+
+
+def find_bounds(image_file: str, text: str, regex: bool = False) -> list[tuple[int, int, int, int]]:
```
ichard26, line 4: I thought this project was using `pathlib.Path` objects for filesystem paths? Please aim for consistency.

itinerant-iterators/app/obfuscate/find_bounds.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    data = {
+        "text": text,
+        "regex": regex
+    }
+    files = {
+        'file': open(image_file, 'rb')
+    }
+    bounds = requests.post("https://itinerantiterators.hipeople21.repl.co", data=data, files=files)
+    return bounds.json()
```
ichard26, line 23: Any sort of networking can easily fail for any number of reasons. The response status code should be checked before returning the data. Additionally, there should be a try... except... where this function is called that catches `requests.HTTPError` and emits a clear error to the user.

```suggestion
    response = requests.post("https://itinerantiterators.hipeople21.repl.co", data=data, files=files)
    response.raise_for_status()
    return response.json()
```

itinerant-iterators/app/obfuscate/find_bounds.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+if __name__ == "__main__":
```
ichard26, line 26: Excellent usage of `if __name__ == "__main__"` convention to provide debugging functionality!

itinerant-iterators/app/obfuscate/obfuscate.py
```diff
@@ -0,0 +1,40 @@
+from typing import Protocol
+
+from PIL.Image import Image as PILImage
+
+
+class Obfuscate(Protocol):
+    """Obfuscate defines the behaviour to obfuscate the part of the image given in the bounding box."""
```
ichard26, line 7: I'm very impressed to see Protocols used! It's a great design pattern for implementing the obfuscators and steganography (algorithms).

One thing about the name though, I think it would be clearer to have the protocol named `Obfuscator` or even `ObfuscationMethod`. Right now, it is a subpackage, module, and class name. That's too many meanings for a single name.

itinerant-iterators/app/obfuscate/obfuscate.py
ichard26, line 1: I would suggest renaming the `app.obfuscate.find_bounds` module to `bounds` and move the `validate_bounding_box()` function to it. In addition, I would move the `Obfuscate` protocol to `__init__.py` (or at least rename this module to be a bit clearer). I'm not convinced this extra layer of organisation is beneficial.

itinerant-iterators/app/steganography/lsb.py
```diff
@@ -0,0 +1,85 @@
...
+        for pixel in img:
+            for channel in pixel:
+                lsb = channel & 1
+                char += (2**bit_idx) * lsb
+                bit_idx = (bit_idx - 1) % 8
+                if bit_idx == 7:
+                    text += chr(char)
+                    char = 0
+                    if text.endswith(EOM):
+                        return text[: -len(EOM)]
```
ichard26, line 85: Include a link to a description of the algorithm implemented here as the bare minimum. In addition, comments explaining how this works would be great.

itinerant-iterators/app/steganography/lsb.py
```diff
@@ -0,0 +1,85 @@
...
+            for idx, channel in enumerate(pixel):
+                try:
+                    bit = next(bits)
+                    if bool(bit):
+                        pixel[idx] = byteutils.set_bit(channel, 0)
+                    else:
+                        pixel[idx] = byteutils.clear_bit(channel, 0)
+                except StopIteration:
+                    img.pixels = pixel_data
+                    return
```
ichard26, line 62: Aim to put the least amount of code within the try body so no exceptions are inadvertently caught. Also, `idx` can be better named.

```suggestion
        for pixel in self.pixels_generator(pixel_data, bytes_input=bytes_input):
            for channel_idx, channel in enumerate(pixel):
                try:
                    bit = next(bits)
                except StopIteration:
                    img.pixels = pixel_data
                    return
                else:
                    if bit:
                        pixel[channel_idx] = byteutils.set_bit(channel, 0)
                    else:
                        pixel[channel_idx] = byteutils.clear_bit(channel, 0)
```

itinerant-iterators/app/steganography/lsb.py
```diff
@@ -0,0 +1,85 @@
...
+        with `Image.save`.
+
+        Args:
+            text (str): text to encode in the image
+            img (Image): Image to encode the text in
+
+        Raises:
+            ValueError: If the input text cannot be encoded in the image.
+        """
+        bytes_input = (text + EOM).encode()
```
ichard26, line 45: Choose a text encoding so differences in system default encodings doesn't cause incompatibilities.

itinerant-iterators/app/obfuscate/obfuscate.py
```diff
@@ -0,0 +1,40 @@
...
+    size = image.size
+    left, up, right, down = box
+    return (
+        left >= 0
+        and right <= size[0]
+        and up >= 0
+        and down <= size[1]
+        and right >= left
+        and down >= up
+    )
```
ichard26, line 40:
```suggestion
    image_length, image_width = image.size
    left, up, right, down = box
    return (
        left >= 0
        and right <= image_length
        and up >= 0
        and down <= image_width
        and right >= left
        and down >= up
    )
```

itinerant-iterators/app/obfuscate/colour_box.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def hide(self, box: tuple[int, int, int, int], image: PILImage):
+        """Hide the pixels in the bounding box given in the image.
+
+        Args:
+            box (tuple[int, int, int, int]): Bounding box to obfuscate
+            image (PILImage): Image to obfuscate
+        """
+        for i in range(box[0], box[2]):
+            for j in range(box[1], box[3]):
```
ichard26, line 39: Are the bounds inclusive? Using range like this may cause an off-by-one error.

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+        Args:
+            b (bytes): bytes to be encoded in the image
+
+        Returns:
+            bool: whether the given bytes can be encoded in the image
+        """
+        return len(b) <= self.max_bytes
+
+    @property
+    def pixels(self) -> list[list[int]]:
```
ichard26, line 43: I understand that Pillow may return a tuple of varying length due to RGB/RGBA images, but using a type alias would still improve readability.

```python
Pixel = list[int]  # Each pixel is a list of 3-4 integers, each representing a channel (RGB/A)

def pixels(self) -> list[Pixel]:
    ...
```

If this application only works with images with 3 or 4 channels, then using a fixed-length tuple type alias would be best.

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+        """Get the pixels of the image
+
+        Returns:
+            list[list[int]]: Array of pixels in the image, where each pixel is a list of integers
+        """
+        return list([list(rgb) for rgb in self.image.getdata()])
```
ichard26, line 49: List comprehensions, as their name suggests, build a new list. No need to call list again.

```suggestion
        return [list(rgb) for rgb in self.image.getdata()]]
```

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @pixels.setter
+    def pixels(self, newdata: list[list[int]]):
+        """Set the pixels of the image
+
+        Args:
+            newdata (list[list[int]]): Array of pixels to set in the image, where each pixel is a list of integers
+        """
+        newdata = list([tuple(rgb) for rgb in newdata])
```
ichard26, line 58: Ditto.
```suggestion
        newdata = [tuple(rgb) for rgb in newdata]
```

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+        """Save the image to the given file
+
+        Args:
+            file (Path): File to save the image to
+        """
+        # Saving as JPEG breaks encoding due to lossy compression.
+        self.image.save(file, format="PNG")
+
+
+class Steganography(Protocol):
```
ichard26, line 71: Like my earlier comment, I think a better name would be `SteganographyAlgorithm`. Wordy I know, but readability is more important (and most editors have autocomplete features these days...)

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+    def can_encode(self, b: bytes) -> bool:
+        """Check whether the given bytes can be encoded in the image
+
+        Args:
+            b (bytes): bytes to be encoded in the image
+
+        Returns:
+            bool: whether the given bytes can be encoded in the image
+        """
+        return len(b) <= self.max_bytes
```
ichard26, line 40: This is unnecessary indirection. The caller can simply reimplement the comparison themselves (which would be clearer anyway IME).

itinerant-iterators/app/steganography/steganography.py
ichard26, line 1: The `Image` class is appropriate used, but I'm not convinced it is worth the implementation complexity. I'm biased so take this with a grain of salt, but a module with the methods turned into utility functions may be simpler.

itinerant-iterators/app/steganography/lsb.py
```diff
@@ -0,0 +1,85 @@
...
+    for pixel in pixels:
+        yield pixel
+
+
+def pixels_space_evenly(pixels: list[list[int]], bytes_input: bytes):
+    """A generator that chooses pixels such that they are spaced evenly throughout the image"""
+    ...  # TODO
+
+
+class Lsb(Steganography):
```
ichard26, line 18: Sometimes, spelling out an acronym in full is better in the longrun.

```suggestion
class LeastSignificantBit(Steganography):
```

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+    "infile",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.argument(
+    "outfile",
+    type=click.Path(exists=False, dir_okay=False, file_okay=True, writable=True),
+)
+def watermark_encode(text: str, infile: Path, outfile: Path):
+    """Entrypoint to watermark encode command"""
+    steganography.encode(text, infile, outfile, steganography.Lsb())
```
ichard26, line 49: This is purely stylistic, but using keyword arguments is often more readable for booleans and other "magic values". 

```suggestion
    steganography.encode(text, infile, outfile, algorithm=steganography.Lsb())
```

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+def watermark_decode(infile: Path):
+    """Entrypoint to watermark decode command"""
+    lsb = steganography.Lsb()
+    result = steganography.decode(infile, lsb)
+    if result:
+        click.echo(result)
+    else:
+        click.echo(f"no encoded data found in {infile}")
```
ichard26, line 64: Programs can set a return code to communicate a general status on completion. A return code of zero usually signifies "all good, nothing went wrong" while a non-zero return code signals some sort of error. 
```suggestion
    if result:
        click.echo(result)
    else:
        click.echo(f"no encoded data found in {infile}")
        sys.exit(1)
```

This is especially important if the CLI is going to be used in an automated fashion.

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Args:
+        image (Image): PIL Image to display
+
+    Returns:
+        io.BytesIO: Buffer of image bytes to display
+    """
+    tmp_img = image
+    # Get image size
+    cur_width, cur_height = image.size
```
ichard26, line 255: While comments are usually a good thing, avoid comments that only state the obvious. They're unnecessary (and can fall out of date with the code). Aim for comments that explain the WHY behind the code it's for (and WHAT if the code is implementing complex logic).

For example, this function seems to resize the image to ensure it stays within a 600x400 size limit, that would be perfect to include within a comment (or docstring).

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+    # If image width or height oversize, resize the image
+    if cur_width > 600 or cur_height > 400:
+        new_width = 600 if cur_width > 600 else cur_width
+        new_height = 400 if cur_height > 400 else cur_height
```
ichard26, line 259:
```suggestion
        new_width = min(600, cur_width)
        new_height = min(400, cur_height)
```

itinerant-iterators/app/cli.py
```diff
@@ -0,0 +1,100 @@
...
+):
+    """Entrypoint to obfuscate command"""
+    # setup obfuscator
+    obfuscator = obfuscate.ColourBox(colour)
+    if mode == "blur":
+        obfuscator = obfuscate.BlurBox()
+    elif mode == "colour":
+        obfuscator = obfuscate.ColourBox(colour)
+
+    img = Image.open(infile)
```
ichard26, line 91: Don't forget to close the image once it isn't needed anymore. Leaving it unclosed can result in these warnings:

```
/home/ichard26/programming/misc/python-discord/repos/compare-work-dir/repos/itinerant-iterators/app/cli.py:96: ResourceWarning: unclosed file 
<_io.BufferedReader name='/home/ichard26/Pictures/Screenshot from 2023-08-12 16-56-52.png'>
  bounds = obfuscate.find_bounds(infile, text, regex)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+            bounds = find_bounds(filename, text, values['is_regex'])
+            img = Image.open(filename)
+            if values["box"] is True:
+                colour_box = ColourBox((0, 0, 0))
+                for bound in bounds:
+                    colour_box.hide(bound, img)
+            elif values['blur'] is True:
+                blur_box = BlurBox()
+                for bound in bounds:
+                    blur_box.hide(bound, img)
```
ichard26, line 239: Select an obfuscator first and then use that in a for loop like the CLI implementation. Having two loops that do the same thing is unnecessary complexity.

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+            bio = _get_image_for_ui(img)
```
ichard26, line 240: Don't forget to update `image` after obfuscation, otherwise saving will save the old, unmodified image.

itinerant-iterators/app/ui.py
ichard26, line 1: The application's GUI works very nicely (barring a few bugs). Well done on using PySimpleGUI!

A note on code structure, I would suggest moving the UI layout definitions to be module level constants to avoid a level of indentation/nesting. I would also like to see the various "actions" to be handled by individual functions, in lieu of having all of the UI code in a single massive function.

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+    WATERMARK_TEXT_INPUT = "-WATERMARK-TEXT-INPUT-"
+    WATERMARK_ENCODING_MODE = "-WATERMARK-ENCODING-MODE-"
+    APPLY_WATERMARK = "-APPLY-WATERMARK-"
+    VIEW_WATERMARK = "-VIEW-WATERMARK-"
+    APPLY_OBFUSCATE = "-APPLY-OBFUSCATE-"
+    OBFUSCATE_TEXT_INPUT = "-OBFUSCATE-TEXT-INPUT-"
+    WATERMARK_FRAME = "-WATERMARK-"
+    OBFUSCATE_FRAME = "-OBFUSCATE-"
+    IMAGE = "-IMAGE-"
+    IMAGE2 = "-IMAGE2-"
```
ichard26, line 25: Enums are underrated IMO, it's great to see them used here.

itinerant-iterators/app/ui.py
```diff
@@ -0,0 +1,267 @@
...
+            bio = _get_image_for_ui(steg_image.image)
+            window[UIKey.IMAGE].update(bio.getvalue())
+        elif event == UIKey.VIEW_WATERMARK:
+            lsb = steganography.Lsb()
+            steg_image = steganography.Image(filename)
+            decoded_text = lsb.decode(steg_image)
+            if decoded_text is None:
+                sg.Popup("Image does not contain any encoded data!")
+            else:
+                window[UIKey.WATERMARK_TEXT_INPUT].Update(decoded_text)
```
ichard26, line 227: So the reason why applying watermark doesn't seem to work is that this code uses the original image (path). The `image` variable is correctly updated after applying the watermark (so saving and then using the CLI to decode works), but that's ignored here.

This is a bit annoying to fix because `steganography.Image` doesn't accept a Pillow `Image` instance. Ideally, you would be able to pass the `image` variable so any updates made to it would be picked up. This is one of the risks of implementing class-based abstractions, they work great if they fit the needs, but are an absolute pain if they don't quite match. Using a set of functions is generally more flexible. But, updating the  `steganography.Image` class to accommodate this usecase wouldn't be a bad option either.

itinerant-iterators/app/byteutils.py
```diff
@@ -0,0 +1,42 @@
...
+def iter_bits(bytes_: bytes) -> Generator[int, None, None]:
+    """Enumerate over each bit in the given bytes.
+
+    Args:
+        bytes_ (bytes): bytes to enumerate over
+
+    Yields:
+        Generator[int, None, None]: Return each bit as an int (0/1)
+    """
+    for byte in bytes_:
```
ichard26, line 13: I'm impressed to see the terminating underscore convention for names that would (otherwise) shadow a built-in name, kudos!

itinerant-iterators/app/steganography/steganography.py
```diff
@@ -0,0 +1,123 @@
...
+    def __iter__(self):
+        """Iterate over the image's pixels."""
+        return iter(self.image.getdata())
+
+    @property
+    def image(self) -> PILImage:
+        """Get the image"""
+        return self.__image
+
+    @cached_property
```
ichard26, line 25: Solid use of `functools.cached_propery`!

ichard26 (commented):
_(Part. 2: Code Review of the application)_

## Code quality & structure

I appreciate and encourage the use of type annotations! It helps massively with the readability and maintainability of the code. However, there are certain places where type annotations are incomplete or lacking. Some APIs expect callables or complex data structures and it'd be better to thoroughly type them.

On top of that, I'd highly suggest using type aliases. Not only do they help with reducing repetition, but they give names to often complex types. For example, a `Bound` type alias for `tuple[int, int, int, int]` would be amazing!

For example, here's the beginning of the `Lsb` class but with the feedback above incorporated. Notice how the type annotations further explain what the code does at a glance, improving readability ("what does pixels_generator do?" ... "ah I know, it takes a list of Pixels and yields pixels in a certain order").

```python
from collections.abc import Iterable
from typing import Callable

# This type alias should be defined somewhere else in practice,
# but for sake of the example, it'll live here.
Pixel = tuple[int, int, int]

class Lsb(Steganography):
    def __init__(
        self, pixels_generator: Callable[[list[Pixel], bytes], Iterable[Pixel]] = pixels_top_left
     ): ...
```

Finally, I forgot to mention this earlier, but I'd firmly encourage using a type checker (eg. [mypy](https://mypy-lang.org/) or [pyright](https://microsoft.github.io/pyright/#/)) to check and enforce types. Using type annotations is one thing, but having machine checked type annotations (and by extension, type documentation) is even better! 

The application code is generally well organised. I like how the obfuscation and steganography code are sectioned off in their own subpackages. Most functions and classes are well designed, naturally documenting and delineating sections of logic (in a way that follows [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) and the [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle)). 

Certain functions and modules could be renamed and moved, but there isn't anything truly terrible. I do think some portions are overengineered and can be simplified (eg. some classes could be replaced with a module with a set of functions). Overall, it was generally easy to read the code and figure out the logic flow across the various modules.

Anyhow, that concludes my overall comments. Please see the comments below for specific feedback.

## Code documentation

I'm pleased to see docstrings. They're consistently used and complete with all parameters and return values documented. However, many arguments and return values could be further specified, and some docstrings only state the obvious. Overall, they are decent and I would rather have them than not, but there is some room to improve. 

More importantly, the codebase would strongly benefit from more comments. There are some areas that implement very involved logic and are challenging to read and follow. Adding comments that explain what is going on, especially the why would be a good next step for improving readability. Cases where this problem was particularly bad are noted in this review's comments.

The README is short, but it's clear and and does a good job at highlighting the features and the project can be used. I'm glad examples and installation docs are included. Kudos!

## User interface (CLI/GUI)

> **Attention**: I have never used PySimpleGUI so I was unable to make specific comments on its use. I tried my best to provide feedback on the other parts though.

The CLI and GUI are well designed and work well, but I would suggest a different way to handle the GUI/CLI distinction. Since the command line interface is (according to the README) the primary way the application is meant to be used, instead of implementing a "launcher"-type CLI, I would suggest placing the GUI start up code in a `ui.py` module guarded by `if __name__ == "__main__":`.

This simplifies the whole CLI, removing a layer of complexity. Instead of commands like `python -m app cli watermark decode image.png`, an user can drop the (clunky) `cli` part and write `python -m app watermark decode image.png`. (All while starting the GUI is still easy: `python -m app.ui`.)

In addition, if the application was packaged using say, [flit](https://flit.pypa.io/en/stable/), so it could be installed using `pip`, [project scripts](https://flit.pypa.io/en/stable/pyproject_toml.html#scripts-section) could be defined so once installed, the user could run (pretend the application is called "imgedit") `imgedit` or `imgedit-gui` in their terminal to launch the CLI/GUI for a smooth terminal UX.

Finally, to bring it back to the CLI, it could do with some better documentation. Command docstrings should include rich descriptions and examples and most options should have a description too. I left a comment on `app/cli.py` that goes into further detail. (TL;DR, it shows when `--help` is passed`)

itinerant-iterators/server/find_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+        list[tuple[int, int, int, int]]: List of bounding boxes for each character to hide
+    """
+    print("Creating a named temporary file..")
+    temp = tempfile.NamedTemporaryFile()
+    os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = temp.name
+
+    temp.write(bytes(json.dumps(dict(os.environ)), 'utf-8'))
+    temp.seek(0)
+    client = vision.ImageAnnotatorClient()
+    temp.close()
```
ichard26, line 28: I would suggest reading the Google API credentials using a different method. Grabbing everything in `os.environ`  and then putting them in a "temporary"[^1] file is not recommendable. Either pass the credentials directly to the client from certain environment variables or require the `GOOGLE_APPLICATION_CREDENTIALS` envvar to be set to a valid credentials file at start up.

[^1]: I'm not sure what tempfile does exactly on shutdown, but there's a decent chance the temporary file is never cleaned up as it's not closed anywhere as far as I can tell.

itinerant-iterators/server/find_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+    response = client.document_text_detection(image=image)
+    document = response.full_text_annotation
+    if (text.lower() not in document.text.lower() and not regex):
+        return []
+
+    if regex:
+        matches = re.findall(text, document.text)
+        if len(matches) > 0:
+            if isinstance(matches[0], tuple):
+                text = list(itertools.chain(*matches))
```
ichard26, line 42: Very cool to see `itertools` be used, but this is a bit more idiomatic. 
```suggestion
                text = list(itertools.chain.from_iterable(matches))
```

itinerant-iterators/server/find_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+                                 symbol.bounding_box.vertices[2].y)
+                            ])
+                        else:
+                            image_chars.append([
+                                symbol.text,
+                                (symbol.bounding_box.vertices[0].x,
+                                 symbol.bounding_box.vertices[0].y,
+                                 symbol.bounding_box.vertices[2].x,
+                                 symbol.bounding_box.vertices[2].y)
+                            ])
```
ichard26, line 66: There's only one minor difference between the two branches, merge them.
```suggestion
    image_chars = []
    for page in document.pages:
        for block in page.blocks:
            for paragraph in block.paragraphs:
                for word in paragraph.words:
                    for symbol in word.symbols:
                        image_chars.append([
                            symbol.text if regex else symbol.text.lower(),
                            (symbol.bounding_box.vertices[0].x,
                             symbol.bounding_box.vertices[0].y,
                             symbol.bounding_box.vertices[2].x,
                             symbol.bounding_box.vertices[2].y)
                        ]
```

itinerant-iterators/server/find_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+    if (text.lower() not in document.text.lower() and not regex):
+        return []
+
+    if regex:
+        matches = re.findall(text, document.text)
+        if len(matches) > 0:
+            if isinstance(matches[0], tuple):
+                text = list(itertools.chain(*matches))
+            else:
+                text = matches
```
ichard26, line 44: Repurposing variables, especially parameter variables, tends to make it harder to keep track of what's happening and how data is flowing/being modified. Modifying a parameter variable is okay, but repurposing it for something else is almost never a good idea. Keep the different pieces of data separate.

itinerant-iterators/server/find_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+                text_index += 1
+                temp_bounds.append(image_char[1])
+                if text_index == len(text):
+                    bounds.extend(temp_bounds.copy())
+                    temp_bounds.clear()
+                    text_index = 0
+            elif text[text_index].lower() != image_char[0].lower():
+                text_index = 0
+                temp_bounds.clear()
+    else:
```
ichard26, line 89: This is frankly hard to follow. I would suggest adding comments and changing variable names to be more descriptive. With simple logic, it's possible to get away with bad variable names, but once there's some complexity, good names and comments are crucial!

> From what I can glean, this seems to go through each character in the image (as detected by GCV) and checks whether it matches a character in the text to filter. If there's a match, then the next character in the image is checked to see whether it matches the next character of the filtered text (while skipping spaces) and so on...
>
> This is a reasonably efficient and straightforward algorithm! (despite the readability issues with the implementation)

(A good start would be including the description I just provided or your own in a comment as a heads-up/aid to  the reader.)

For example, this is a rewrite of this code with better names and with a different approach to keeping track of what characters of the text we're looking for have been found:

```python
    if not regex:
        search_characters = list(text.lower().replace(" ", ""))
        remaining_search_characters = search_characters.copy()
        current_matching_bounds = []
        for image_char, image_char_bound in image_chars:
            # Grab the next character we're still looking for.
            try:
                search_char = remaining_search_characters.pop(0)
            except IndexError:
                # No more characters are left to check. Restart.
                remaining_search_characters = search_characters.copy()
                current_matching_bounds.clear()

            if search_char == image_char.lower():
                current_matching_bounds.append(image_char_bound)
                if len(remaining_search_characters) == 0:
                    bounds.extend(current_matching_bounds)
                    remaining_search_characters = search_characters.copy()
                    current_matching_bounds.clear()
            else:
                # A character did not match, bail and reset the search attempt
                remaining_search_characters = search_characters.copy()
                current_matching_bounds.clear()
```

I did not test this, so this may contain bugs and may not produce the exact same results, but the idea behind my feedback is there.

ichard26 (commented):
_(Part. 3: Code Review of the server)_

It's neat how your project incorporates Google Cloud Vision. I like how regex is used here to build a really cool feature (regex-based obfuscation). However, the server-side code needs improvement. As-is, it's hard to read and is dubiously written. Other than my specific comments, my two big suggestions are to:

- Break out logical sections into individual functions. This will encourage code reuse (and there's a fair of fairly repetitive code here) and cleaner variable use (several variables have multiple purposes which is bad). Smaller functions tend to be also easier to read since there's just less going on.[^1] 
- Write more comments, I really mean it :)

[^1]: It's a balancing act though. Sometimes too many functions is a bad thing. It's like blank lines in an essay. You shouldn't write your whole essay without a single blank line, but you also don't want to have a blank line for every 4 sentences. A general rule of thumb is each function should implement a single "step" of your program's _overall_ logic (the little details usually don't matter/warrant their own tiny functions).

itinerant-iterators/tests/test_byteutils.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # act
+    gen = byteutils.iter_bits("a".encode())
+
+    # assert
+    for want_bit in want_bits:
+        assert next(gen) == want_bit
+
+    with pytest.raises(StopIteration):
+        next(gen)
```
ichard26, line 19: This is great! I love how a testing format is clearly being followed and how the API contract of `byteutils.iter_bits()` is being tested.

itinerant-iterators/tests/test_byteutils.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_set_bit():
+    cases = [
+        (4, 0, 5),
+        (5, 0, 5),
+        (2, 3, 10),
+    ]
+    for case in cases:
+        got = byteutils.set_bit(case[0], case[1])
+    assert got == case[2]
```
ichard26, line 30: How about this? (the assert was also supposed to be per-loop, right?)
```suggestion
    for value, position, expected in cases:
        got = byteutils.set_bit(value, position)
        assert got == expected
```

itinerant-iterators/tests/test_byteutils.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_clear_bit():
+    cases = [(4, 0, 4), (5, 0, 4), (10, 3, 2)]
+    for case in cases:
+        got = byteutils.clear_bit(case[0], case[1])
+        assert got == case[2]
```
ichard26, line 37: The same unpacking assignment trick used above works here too.

itinerant-iterators/tests/test_steganography.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_lsb_encode_decode():
+    message = "secret message"
+    lsb = Lsb()
+    f = tempfile.NamedTemporaryFile("w", delete=True, suffix=".png")
+    encode(message, Path("tests/data/python-logo.jpg"), Path(f.name), lsb)
+    got = decode(Path(f.name), lsb)
+    f.close()
+    assert message == got
```
ichard26, line 15: Solid integration test. Could be made cleaner though.
```suggestion
def test_lsb_encode_decode():
    message = "secret message"
    with tempfile.NamedTemporaryFile("w", delete=True, suffix=".png") as _f:
        temporary_file = Path(_f)
        encode(message, Path("tests/data/python-logo.jpg"), temporary_file, algorithm=Lsb())
        decoded_message = decode(temporary_file, lsb)
    assert message == decoded_message
```

If you don't understand what the `with` statement is, [here's a Real Python link](https://realpython.com/python-with-statement/), but anyway, it just manages the cleanup of the temporary file. Once execution leaves the with block, the temporary file is closed (and deleted).

itinerant-iterators/tests/test_obfuscate.py
```diff
@@ -0,0 +1,114 @@
...
+        elif pixel == (255, 255, 255):
+            white += 1
+        else:
+            raise Exception(
+                f"obfuscated image contained pixel of unexpected colour: {pixel}"
+            )
+
+    assert img.size == (100, 100)
+    assert black == 2500
+    assert white == 7500
```
ichard26, line 66: This seem like a great candidate for [parametrized tests](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html). 

```python
@pytest.mark.parametrize(
    "obfuscator", [ColourBox("black"), ColourBox((0, 0, 0))], ids=["by_name", "by_tuple"]
)
def test_obfuscate_colour_box(obfuscator):
    # arrange
    # create a new blank image with all white pixels
    img = Image.new("RGB", (100, 100), (255, 255, 255))

    # act
    obfuscator.hide((25, 25, 75, 75), img)

    # assert
    all_pixels = list(img.getdata())
    white = 0
    black = 0
    for pixel in all_pixels:
        if pixel == (0, 0, 0):
            black += 1
        elif pixel == (255, 255, 255):
            white += 1
        else:
            raise Exception(
                f"obfuscated image contained pixel of unexpected colour: {pixel}"
            )

    assert img.size == (100, 100)
    assert black == 2500
    assert white == 7500
```
Also, perhaps the output could be checked against a known-good image file? That's real nickpicky though :p

ichard26 (commented):
_(Part 4. Code review of the tests)_

Overall well done with the tests! It's great to see tests (and well designed tests!) included in the project, they're an excellent way to up and maintain code quality even in the face of rapid changes. I just have a few suggestions to make the code quality even better.

ichard26:
_(Part 5: Final comments :tada:)_

Your team's project is on the smaller side, but it felt complete and generally polished (barring a few bugs with the GUI). I enjoyed using your application during this code jam! The code quality is decent overall. There were many things I liked (eg. tests, the use of `typing.Protocol`, the use of Click), but like almost anything in life, there was also room to improve. Summarizing from my lengthy review, the main suggestions are:

- Use better names for variables and classes
- Use functions to organise code into logical sections[^1]
- Adopt a type checker to enforce quality, reliable, and complete type annotations
- And finally, more comments please! I sound like a broken record on this point, but I really mean it..

Please don't take my review as an attack on your abilities — despite my many comments, I'm pleased with your submission — we just want to provide y'all with constructive and detailed feedback so you can all improve. :duck: 

If you have any questions or want to discuss my feedback, please reach out! Also, I'm far from being a perfect developer, so if you disagree with some of my feedback, that's okay! Take what you want from this review.

Congratulations on making Top 10. Good luck for the livestream!

[^1]: this basically only applies to the server code!

PR #7, opened 2023-09-28
Readable Regexes

janine9vn:


readable-regexes/README.md
```diff
@@ -0,0 +1,43 @@
+[![Python tests](https://github.com/Artemis21/pydis-jam23/actions/workflows/test.yaml/badge.svg?branch=main)](https://github.com/Artemis21/pydis-jam23/actions/workflows/test.yaml)
+
+# Readable Regexes: Stego Suite
+
+![A screenshot of the app home page, showing a list of codecs to choose from alongside an image of the app logo](https://github.com/Artemis21/pydis-jam23/assets/57376638/bbf37666-b9e6-4701-bf1a-70f6e0a73ec5)
+
+Stego Suite is a toolkit providing a variety of approaches to [steganography](https://en.wikipedia.org/wiki/Steganography), developed as part of [the Python Discord Code Jam, 2023](https://www.pythondiscord.com/events/code-jams/10/).
+
+Watch this [video](https://youtu.be/p08M0LqjWMI) to get an overview of the project.
```
KeithTheEE, line 9: There should be a description summarizing the project here as well. The video should act as a supplement or a chance for more in-depth usage, and not watching the video shouldn't stop a user from understand what a project does.

readable-regexes/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+[The wiki](https://github.com/Artemis21/pydis-jam23/wiki) contains instructions for [installation and development](https://github.com/Artemis21/pydis-jam23/wiki/Installation), [an introductory guide](https://github.com/Artemis21/pydis-jam23/wiki/Guide) and information on [the available codecs](https://github.com/Artemis21/pydis-jam23/wiki/Codecs).
```
KeithTheEE, line 11: The wiki is really good content. I do wish some of it was in the readme, but it's well done and very helpful.

readable-regexes/src/pydis_jam23/cli_app.py
```diff
@@ -0,0 +1,109 @@
+import argparse
```
KeithTheEE, line 1: argparse is one of my favorite modules, I'm always happy when I see it used

readable-regexes/src/pydis_jam23/cli_app.py
```diff
@@ -0,0 +1,109 @@
...
+        for param in codec.encode_params:
+            add_codec_arg(parser, codec, param, when="encode")
+        for param in codec.decode_params:
+            add_codec_arg(parser, codec, param, when="decode")
+    return parser
+
+
+def add_codec_arg(parser: argparse.ArgumentParser, codec: Codec, param: CodecParam, when: str) -> None:
+    settings: dict[str, typing.Any]
+    if issubclass(param.type_, bool):
```
KeithTheEE, line 67: I unfortunately don't use match case statements, but this structure feels perfect for a match case statement. You're using 3.11 for this project, take advantage of the latest changes in the python language and use projects like these as a chance to learn about those changes.

readable-regexes/src/pydis_jam23/codecs/__init__.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Protocol
+
+from PIL import Image
+
+from . import concat, edges, lsb, noise, notlsb, ssdb
+
+from .common import CodecError, CodecParam
+
+
+class Codec(Protocol):
```
KeithTheEE, line 10: I am not overly thrilled by this structure, but I think most of that comes from the absence of comments explaining what is and (more importantly) is not being done in this class. There's also some functions in the sub modules which are repeated, and should probably be in this init or the common file.

readable-regexes/src/pydis_jam23/codecs/common.py
```diff
@@ -0,0 +1,51 @@
+import pathlib
+from collections.abc import Callable
+from dataclasses import dataclass
+
+
+class CodecError(Exception):
+    """An error encountered while trying to perform message encoding/decoding."""
+
```
KeithTheEE, line 8: The `__init__.py` and the `common.py` file both seem to achieve the same purpose--starting an encode/decode algorithm, and hold the basic functions that the encode and decode algorithms need. If every codec needs the init and the common functions, then the contents should be in the same file (or grouped somehow--as a project grows splitting them into two files makes sense, but those should both be isolated in their own directory).

readable-regexes/src/pydis_jam23/codecs/common.py
```diff
@@ -0,0 +1,51 @@
...
+
+@dataclass
```
KeithTheEE, line 10: Great use of dataclass, I constantly see it and think that I need to learn how to fully utilize it and seeing it here both makes sense for what's being done, and makes me want to start using it.

readable-regexes/src/pydis_jam23/codecs/common.py
```diff
@@ -0,0 +1,51 @@
...
+    required: bool
+
+    display_name: str
+    help_: str
+    cli_flag: str
+
+
+ASSETS = pathlib.Path(__file__).parent.parent / "assets"
+SEVEN_BIT_MAX = 127
+
```
KeithTheEE, line 24: Bit masks are a concept that would be exceptionally useful, and useful in this common file. 
```py
mask_1   = 0b0000_0001
mask_2   = 0b0000_0010
mask_4   = 0b0000_0100
...
mask_128 = 0b1000_0000
```

To get any given bit you just take the byte you're evaluating, `x`, and 'and' it  with the corresponding mask
`bit_of_interest = x & mask_4`

If the value is not 0, then the bit was a 1, otherwise it was a 0. Cleverly organized 'and' operations are much faster than bit shifting, so not only would a collection of masks be helpful, it would optimize a lot of the codec processes.

readable-regexes/src/pydis_jam23/codecs/edges.py
```diff
@@ -0,0 +1,263 @@
...
+def encode(image: Image.Image, message: bytes, *, test_channel: int | None = None, **codec_args: Any) -> Image.Image:
+    """encode a message into an image using the above described method"""
+    if codec_args:
+        msg = f"Unexpected codec arguments: {codec_args}"
+        raise TypeError(msg)
+
+    image_data = bytearray(image.tobytes())
+    data = encode_varint(len(message)) + message
+
+    if image.mode not in ALLOWED_MODES.keys():
```
KeithTheEE, line 40: It feels odd that this error is being caught here and not closer to the user interface layer.

readable-regexes/src/pydis_jam23/codecs/edges.py
```diff
@@ -0,0 +1,263 @@
...
+        raise CodecError(msg)
+
+    num_channels = ALLOWED_MODES[image.mode]
+    num_data_channels = num_channels - 1
+
+    if test_channel is None:
+        mask_color = randint(0, num_data_channels)  # noqa: S311 no need for crypto
+    else:
+        mask_color = test_channel
+    alternative_trys = 0
```
KeithTheEE, line 51: The `alternative_trys` check should be broken out into it's own function, not under encode. It's a unique enough operation that it makes sense to section it off so it's easier to understand and modify. Breaking it into it's own function also makes it wildly easier to test, and to understand what exactly is causing an error.

readable-regexes/src/pydis_jam23/codecs/edges.py
```diff
@@ -0,0 +1,263 @@
...
+
+    # set the LSB of the color that is used as information mask to 1
+    image_data[mask_color] = set_lsb(image_data[mask_color], 1)
+
+    # get data spaces at index 1 -> as index 0 (pixel 0,0) marks the color layer used as mask
+    data_indices = generate_data_indeces(edges, mask_color, len(data) * 8, 1, num_channels)
+
+    # split bytes to bits
+    message_binary = []
+    for byte in data:
```
KeithTheEE, line 76: Similar to above, a lot of this should be broken out into it's own function so it's easier to test and to isolate bugs

readable-regexes/src/pydis_jam23/codecs/edges.py
```diff
@@ -0,0 +1,263 @@
...
+    for pixel in image:
+        if pixel == color:
+            count += 1
+    return count
+
+
+def read_next_byte(
+    edges: bytes, image_data: bytes, mask_color: int, pixel_index_offset: int, channel_count: int
+) -> tuple[bytes, int]:
+    """Read the next byte of data out of an image while accounting for the edges mask
```
KeithTheEE, line 149: This is a very helpful docstring, it tells me what the function needs (and what that means) and tells me what to expect it to return, and also tells me what to expect when things go wrong. 

When things go wrong is going to be the time I read this code the most, so that is fantastic to see, it helps me frame what I should expect from the code before I dig into it. (And because I'm most likely to read code when something's broken, it's extra nice to have comments and docstrings to help me when I'm going to be frustrated and not necessarily thinking clearly)

readable-regexes/src/pydis_jam23/codecs/notlsb.py
```diff
@@ -0,0 +1,66 @@
+"""Encode and decode functions for the "not" codec.
+
+Unlike the regular LSB encoding which encodes a message within a file,
+this non-serious approach takes the *image* and encodes it in another
+image containing text of your secret message. In short, it does the opposite
+of the standard LSB codec.
```
KeithTheEE, line 6: I really like that there's a fun joke project included, thank you for that :)

readable-regexes/src/pydis_jam23/codecs/ssdb.py
```diff
@@ -0,0 +1,152 @@
...
+    if kwargs:
+        msg = f"Unexpected arguments: {', '.join(kwargs)}"
+        raise TypeError(msg)
+    if password is None:
+        msg = "A password is required."
+        raise CodecError(msg)
+    return password
+
+
+def get_bits(data: bytes) -> list[int]:
```
KeithTheEE, line 134: Using bit masks here will be faster than looping through each bit with bitshifting. This is really unoptimized, but it does work. And given the time constraints this choice makes sense. That said, going forward I'd suggest playing with optimization here.

readable-regexes/tests/test_codec_lsb.py
```diff
@@ -0,0 +1,25 @@
...
+        b"Hello, world!",
+        bytes(range(256)),
+    ],
+)
+@pytest.mark.parametrize("bits", [1, 3, 8])
+@pytest.mark.parametrize("msb", [False, True])
+def test_message_roundtrip(wikimedia_image: Image.Image, message: bytes, bits: int, msb: bool) -> None:
+    encoded = lsb.encode(wikimedia_image, message, bits=bits, msb=msb)
+    assert lsb.decode(encoded, bits=bits, msb=msb) == message
+
```
KeithTheEE, line 21: Mostly as a general comment, you can use the unoptimized functions in the testing suite. That way once you get the encoding/decoding, or messy difficult code working, you're able to use the first pass to validate and verify the optimization passes and refactoring.

readable-regexes/tests/test_codec_lsb.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+from PIL import Image
+from pydis_jam23.codecs import CodecError, lsb
+
+from .common import wikimedia_image  # noqa: F401 - import for fixtures
+
+
```
KeithTheEE, line 7: I'd like to see more of an explanation of what's going on in each of the tests. What is being tested, and what does a pass/failure mean. Comments in the code are a great way to make that kind of note.

readable-regexes/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Quick installation guide:
+Make sure you are running at least python3.11.
+```shell
+git clone https://github.com/Artemis21/pydis-jam23
+cd pydis-jam23
+pip install hatch
+```
+
+## Quick execution guide:
```
KeithTheEE, line 21: I am left with a question on the readme that I don't feel is answered:

- What image types can this program process?

KeithTheEE (commented):
# Overall Project Quality

The project is pretty good overall. It could do with a pass of refactoring to organize everything, but for a short timeline it establishes a lot. Very well done all!


When dealing with bit manipulation there's some sections that could do with a lot of optimization--stepping through a bit at a time isn't the most efficient method. That said, it does work as a proof of concept when building out code. A nice trick is to take the unoptimized code and turn it into a test function while writing the optimized function, so you have something to check against. At some point that unoptimized test will probably need to be removed or improved, but optimization can be a major pain and having a known 'truth' version of the function makes it a lot easier. This gets you faster code and a great test at the same time.

The other thing this project could do with is more docstrings and comments, especially near the messy bits that took multiple commits to get right. Code is read more often than it's written, and when something takes a while to get right, it'll probably take a while to read and understand. Comments at those points add clarity, and make it easier to track down what broke when a change causes a surprise failure.

 

## Readme

The readme successfully establishes what the program is (A suite for Steganography tools), and establishes how to install and run the code, but does not take the next step to show what this toolkit can do. It links to a video and wiki, but the readme should leave me with a sense of what the project does so I can get a sense of if I should set aside time to try to figure out how to use it. 

Linking to the Wiki is a great idea for more in depth tutorials, guides, and showcasing, but the readme should have something which establishes some of what the project can do, and the breadth of things the project can do. Your readme is going to be the first and maybe last thing a lot of people see when looking over your project. If they don't know what they can do with it after reading the readme, they very well might just search for a different program that tells them up front rather than digging into wiki pages. 

The wiki itself is very verbose. I'm really glad the wiki is here, it's a well written explanation of what was going on under the hood of the program. 


# Code Quality Notes

Overall the code is pretty well laid out. It could use some refactoring now that the codecs have grown, and the user interfaces should be grouped together, but as is it's very functional. 

A couple of areas this project is lacking in:

There needs to be comments in the code. Docstrings, an explanation of what kind things might cause problems, flags for the next time a person reads the code. There were a few inconsistently present, and trying to read through the code was a bit more difficult with few or no comments. I'd wonder if I was correctly understanding what the code was doing. Comments really help with that. 

Add more checks on the user inputs--you can address a lot of common errors by checking if they provided the correct file type, etc. and just save a lot of processing time on it that way. Waiting until deep in the codec to raise the error can make debugging difficult, make it harder to duplicate the same kind of error, and make the total kinds of errors a program can throw messier. 

There will be codec errors which should be raised from the codec, but user input errors shouldn't be allowed to survive that long. 

## Codecs

When making bit masks, I see 
`SEVEN_BIT_MAX = 127`
`byte & 0b0111_1111`
`buf_size = 2**8`

etc. (Buffer size isn't a bitmask, but it still outlines a set of bits for a byte)

Pick a single style, and stick with it throughout the program. When first writing out code and figuring out how it works, using the easiest mask is a great idea, but after a point it makes more sense from a style point of view to approach everything in the same style. Every time you need a set of bits in a given pattern, using only one style, like `x = 0b1111_0000` makes it very visually easy to see what that block of code is doing because the assignment is a constant style--you know this is manipulating data at a bit level. 

`get_bits` should not be handled this way. It's going to be really really slow. You should instead keep a bit index tracker, and just step forward in the bytestream with that. A set of bitmasks make it easy to select the bits of interest. 


## User Interfaces

I would have liked to see notes, comments, and function names on the `web.__init__.py` and the `cli_app.py` file showing them as two boilerplates to the same internals. It makes it easier to manage, track, and debug a program as it grows and a feature is rolled out to one portion, and needs parity in the other. 

## Assets
Keep all util/asset code and files in the same location. CLI assests like fonts, and site assets like logo's are all assets and should be kept together to help with organization and unified style (It's a lot harder to end up with different fonts if the font files are all in the same directory).

## Tests
The tests cover LSB and the Edge encoding method, but the user interface and other codecs are untouched. 

Additionally, there's no comments making it clear what the test achieves and the coverage of the tests appears to be a minimal smoke test. A test exists for a reason: are you just making a smoke test to validate the pipeline, or was a function a pain to get correct and is it really really easy to mess up when updating the function? What does a failed test mean to the developer? Did they mess up a constant or is a function no longer returning a required variable? Comments explaining what the test's passing and failure means can reduce debug time when tests break. 

Tests are hard to do on a tight timeline, but if you want to continue working on this project (because it is fun, especially the `notlsb.py` file) I'd recommend adding more tests, they'll help with the fact that this code is a bit old and unfamiliar. 


# Commit Quality 
Overall it could do with more verbose commit messages. There's bullet points for what was done, but I'm having a hard time getting a sense of why some things were done, especially around some of the codec decisions. This could be supplemented with comments in the code as well. 

The messages do give me an idea of what to expect when looking at the code, while they're not overly descriptive they do highlight what changed which is fantastic. 


# It's been a while since you wrote this code
Take some time to read through it, how much do you remember? Which parts are easy to understand and which parts aren't? The parts that are kind of weird, unfamiliar, or not overly clear as to what they do are the parts that need comments and tests the most. 

There's also a lot of really great blocks of code--take note of those too so you can remember to use that style in future project, and take extra note of "Why" that code stands out to you now. 

This was a fun project to review, thank you for participating in the Code Jam!

readable-regexes/src/pydis_jam23/web/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+from PIL import Image
+
+from pydis_jam23.codecs import CODECS, Codec, CodecError, CodecParam
+
+app = Flask(__name__)
+app.secret_key = "secret"  # We're only running locally.
+
+# Yep, it's a global mutable variable. It *should* work fine though since
+# there's only one user.
+current_image: Image.Image | None = None
```
janine9vn, line 15: Flask provides a `g object` to specifically help deal with this. You can read about it [here](https://flask.palletsprojects.com/en/2.3.x/appcontext/#storing-data).
It's essentially a better way to handle dynamic data that need to stay for the lifetime of the app.

Better yet, I would use a server-side Flask session so that you can support more than a single user. The [Flask-Session](https://flask-session.readthedocs.io/en/latest/) extension is what you would use, usually in combination with something like redis.

readable-regexes/src/pydis_jam23/web/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    global current_image  # noqa: PLW0603 - global mutable
+    if current_image is None:
+        msg = "No image loaded"
+        raise ValueError(msg)
+    codec_obj = find_codec(codec)
+    message = request.form["message"].encode("utf-8")
+    extra_args = get_args(codec_obj.params + codec_obj.encode_params)
+    try:
+        current_image = codec_obj.encode(current_image, message, **extra_args)
+        print(current_image)
```
janine9vn, line 49: This would be a good opportunity to use logging, especially for a server-side process like Flask.

readable-regexes/src/pydis_jam23/web/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+def get_encode(codec: str):
+    return render("action.j2", codec=find_codec(codec), encode=True)
+
+
+@app.route("/encode/<codec>", methods=["POST"])
+def post_encode(codec: str):
+    global current_image  # noqa: PLW0603 - global mutable
+    if current_image is None:
+        msg = "No image loaded"
+        raise ValueError(msg)
```
janine9vn, line 43: I don't think it's valuable to store the message in a separate variable here. It adds an additional line for no particular reason.

janine9vn (commented):
I'm taking a look at the Flask code specifically, so my comments are focused on that.

My two biggest comments are the global variable usage (Flask has a better way of handling that) and also the general project structure. It's a bit odd to see the entire flask project in the `__init__.py`. I would expect to see a more typical directory structure. Even if it doesn't make a functional difference, a more standard code organization makes the code easier for others to navigate and reason about.

I didn't really review the JS or CSS as it's out of scope for our reviews, but it's nice to see following CSS standards with naming conventions and the like. I wanted to shout that out.

PR #9, opened 2023-09-28
Thick Wrappers

janine9vn:


thick-wrappers/helper/decrypt.py
```diff
@@ -0,0 +1,57 @@
...
+    # Iterates through each pixel's binary values and concatenates the least significant bits into decoded code
+    for pixel in RGB_binary_list:
+        output = "0"
+        r, g, b = pixel[0][-N_PLANES:], pixel[1][-N_PLANES:], pixel[2][-1]
+        output += f"{b}{g}{r}"
+        pixel_list.append(int(output, 2))
+
+    # Iterates through each binary output and converts each binary to ASCII, simultaneously checking for the delimiter
+    for binary in pixel_list:
+        if word_total[-5:] == ",,,..":
```
swfarnsworth, line 32: I would expect `',,,..'` to be defined once as a constant, so that you can change the delimiter in one place without having to make any other changes elsewhere in the code base.

You could also have used the `str.endswith` method.

thick-wrappers/helper/decrypt.py
```diff
@@ -0,0 +1,57 @@
...
+            delimiter = True
+            break
+        word_total += format(binary, "c")
+    return word_total[:-5], delimiter
```
swfarnsworth, line 36: Likewise, `word_total[:-5]` can be rewritten with a call to `str.removesuffix`, so that the code doesn't depend on a known length for the delimiter. It might look something like `word_total.removesuffix(DELIMITER)`.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+from helper.constant import N_PLANES
+
+from .constant import (
+    DESCRIPTION, EXIF_MAKE, SOFTWARE_TITLE, STARTING_X, STARTING_Y,
+    TEAM_MEMBERS, Direction, ExifData, ResizeMode, Sizing
+)
+
+T = TypeVar("T", int, np.signedinteger[Any])
+
+EXIF_MODEL_PATTERN = re.compile(r"I(?P<width>\d*)P(?P<height>\d*)P")
```
swfarnsworth, line 19: I love to see regular expressions with named groups 😄

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    strings representing the bits of each pixel's
+    red, green, and blue values.
+    """
+    pixels = img.load()
+    width, height = img.size
+    pixellist = []
+
+    # Iterates through all of the picture's pixels, left to right then down
+    for y in range(height):
+        for x in range(width):
```
swfarnsworth, line 37: This nested loop could be flattened with `itertools.product`

```py
for y, x in product(range(height), range(width)):
    ...
```

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+END_TEXT = ",,,.."
+END_BYTES = list(map(ord, END_TEXT))
+
+
+def encrypt_text_to_image(text: str, image: Image.Image) -> Image.Image | None:
+    """
+    Encode a text string in the pixels of an image
+
+    This function encrypts a string in an image. Non-ASCII characters are
+    stripped from the string, and a copy of the image is made.
```
swfarnsworth, line 17: I recommend this library for converting unicode-only characters to their nearest ascii equivalent: https://pypi.org/project/Unidecode/

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+    if n > extent:
+        return None
+
+    # Create array from ASCII codes with image's width and height,
+    # padded with zeroes
+    bytes = np.array(bytes, dtype=np.uint8)
+    pixels = np.array(image, dtype=np.uint8)
+    mask = np.concatenate(
+        (bytes, np.zeros(extent - n, dtype=np.uint8)), dtype=np.uint8
+    ).reshape((rows, cols))
```
swfarnsworth, line 50: This is a popular way of writing chained method calls in the data science community:
```py
mask = (
    np.concatenate(
        (bytes, np.zeros(extent - n, dtype=np.uint8)), dtype=np.uint8
    )
    .reshape((rows, cols))
)
```
Though what you wrote is fine.

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+
+    modulus = 2 ** N_PLANES
+    # Needed to identify which pixels need LSBs cleared
+    full_rows = n // cols
+    last_row_cols = n % cols
```
swfarnsworth, line 55: This could be written as `full_rows, last_row_cols = divmod(n, cols)`

`divmod(a, b)` returns the tuple `(a // b, a % b)`.

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # First clear all rows where all pixels are used
+    pixels[:full_rows, :, :N_PLANES] = utility.clear_least_significant_bits(
+        pixels[:full_rows, :, :N_PLANES], N_PLANES
+    )
```
swfarnsworth, line 60: Nice work with multi-dimensional indexing here and through the rest of the function.

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+    # Add array to each channel to encode bits
+    # Red
+    pixels[:, :, 0] += mask % modulus
+    mask >>= N_PLANES
+    # Green
+    pixels[:, :, 1] += mask % modulus
+    mask >>= N_PLANES
+    # Blue
+    pixels[:, :, 2] += mask
```
swfarnsworth, line 74: Perhaps you could have used an int enum for the RGB dimension, so that the comments wouldn't be necessary.

Nice work implementing the algorithm to encode the message.

thick-wrappers/helper/encrypt.py
```diff
@@ -0,0 +1,102 @@
...
+    return Image.fromarray(pixels)
+
+
+def encrypt_image_to_image(cover: Image.Image, secret: Image.Image) -> Image.Image:
```
swfarnsworth, line 78: This function looks great. Nice use of type hinting, nice docstring. No suggestions for improvement here.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
+import re
+from glob import glob
+from os.path import join
```
swfarnsworth, line 3: I would have preferred to see `pathlib.Path`, which would eliminate the need for both of these.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    red, green, and blue values.
+    """
+    pixels = img.load()
+    width, height = img.size
+    pixellist = []
+
+    # Iterates through all of the picture's pixels, left to right then down
+    for y in range(height):
+        for x in range(width):
+            r, g, b = pixels[x, y][:N_PLANES]
```
swfarnsworth, line 38: It looks like you're not performing multi-dimensional slicing correctly. Is there a reason you didn't do `pixels[x, y, :N_PLANES]` here?

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+            r, g, b = (
+                str(bin(r))[2:].zfill(8),
+                str(bin(g))[2:].zfill(8),
+                str(bin(b))[2:].zfill(8),
+            )
+            pixellist.append((r, g, b))
+    return pixellist
+
+
+def clear_least_significant_bits(bits: T, n: int) -> T:
```
swfarnsworth, line 48: This function looks good. It will be even better with 3.12's new generic syntax.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    :param image_array: Image numpy array value
+    :param direction: Left bit shit or Right bit shift
+    :param bit_amount: Amount of bits to shift
+    :return: Image data as array
+    """
+    match direction:
+        case Direction.LEFT:
+            return np.left_shift(image_array, bit_amount)
+        case Direction.RIGHT:
+            return np.right_shift(image_array, bit_amount)
```
swfarnsworth, line 74: Nice use of pattern matching. Perhaps you could raise an exception if someone passes one of the other two values for Direction.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def exif_embed_ipp(image_exif: Exif, data: tuple[int, int]) -> Exif:
+    """
+    Embeds In Plain Pixel metadata in an image's Exif data.
+
+    :param image_exif: Image metadata
+    :param data: New metadata to embed
```
swfarnsworth, line 82: This function looks good, except that I'm not clear on what `data` is. Is there a name that might have been more clear?

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+from helper.constant import ResizeMode
+from helper.decrypt import decrypt_image_from_image, decrypt_text_from_image
+from helper.encrypt import encrypt_image_to_image, encrypt_text_to_image
+from helper.utility import exif_embed_ipp, image_resize
+
+InvalidFileError = (OSError, UnidentifiedImageError)
+
+
+@dataclass
+class Filepaths:
```
swfarnsworth, line 17: This class could be greatly improved using pathlib. Though even with this implementation, I would have preferred that the methods not all start with `get` and have inconsistent names (they end with `_fp`, `_output_path`, and `_file_path`).

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+        """Get file path for encrypted output image"""
+        return os.path.join("static", f"output.{self.cover_image_fe}")
+
+    def get_decrypted_output_file_path(self, text=False):
+        """Get file path for decrypted output, either an image or text file"""
+        if text:
+            return os.path.join("static", "output.txt")
+        else:
+            return os.path.join("static", f"output.{self.cover_image_fe}")
+
```
swfarnsworth, line 46: This is how I would have written the class

```py
@dataclass
class FilePaths:
    """Class that has file paths of user image, cover image, and output files"""

    user_text_fe = "txt"
    user_image_fe = ""
    cover_image_fe = ""

    @property
    def text(self):
        """Path for user upload text"""
        return pathlib.Path("static") / f"user_text.{self.user_text_fe}"

    @property
    def user_image(self):
        """Path for user image"""
        return pathlib.Path("static") / f"user_image.{self.user_image_fe}"

    @property
    def cover_image(self):
        """Path for cover image"""
        return pathlib.Path("static") / f"cover_image.{self.cover_image_fe}"

    @property
    def encrypted_image(self):
        """Path for encrypted output image"""
        return pathlib.Path("static") / f"output.{self.cover_image_fe}"

    @property
    def decrypted_image(self):
        return pathlib.Path("static") / f"output.{self.cover_image_fe}"
    
    @property
    def decrypted_text(self):
        """Path for decrypted output text"""
        return pathlib.Path("static") / "output.txt"
```

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+    def get_encrypted_image_output_path(self):
+        """Get file path for encrypted output image"""
+        return os.path.join("static", f"output.{self.cover_image_fe}")
+
+    def get_decrypted_output_file_path(self, text=False):
+        """Get file path for decrypted output, either an image or text file"""
+        if text:
+            return os.path.join("static", "output.txt")
+        else:
+            return os.path.join("static", f"output.{self.cover_image_fe}")
```
swfarnsworth, line 45: It's not clear why this is one method instead of two.

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+        with ui.dialog() as dialog, ui.card():
+            ui.label("Image Output").tailwind(styles.prompt_text_v)
+            ui.markdown(f"![output]({image_output_fp})")
+            with ui.row():
+                ui.button("Download", on_click=lambda: ui.download(image_output_fp))
+                ui.button("Close", on_click=dialog.close)
+        dialog.open()
+    elif os.path.exists(text_output_fp):
+        with open(text_output_fp, "r") as o:
+            text = o.read()
```
swfarnsworth, line 84: This section could be much cleaner with my implementation of the FilePaths class.

```py
if file_paths.text.exists():
    text = file_paths.text.read_text()
```

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+        bytes = f.read()
+    try:
+        text = bytes.decode("utf-8")
+    except UnicodeDecodeError:
+        ui.notify("File could not be read correctly")
+        text_upload.reset()
+        return
+    # Write to storage file
+    with open(file_paths.get_user_text_fp(), "w") as f:
+        f.write(text)
```
swfarnsworth, line 108: This could just be `file_paths.text.write_text(text)`. the `write_text` method would handle opening and closing the file.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    image_exif[ExifData.MAKE.value] = EXIF_MAKE
+    image_exif[ExifData.MODEL.value] = exif_model_builder(data)
+    image_exif[ExifData.ARTIST.value] = TEAM_MEMBERS
+    image_exif[ExifData.SOFTWARE.value] = SOFTWARE_TITLE
+    image_exif[ExifData.DESCRIPTION.value] = DESCRIPTION
+    return image_exif
+
+
+def exif_model_builder(size: tuple[int, int]) -> str:
```
swfarnsworth, line 93: Is this function ever used except the one time in `exif_embed_ipp`? If not, I wouldn't have created it, and just had the logic in the other function. That might help make the other function more explicit.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+                    )
+        case ResizeMode.SHRINK_TO_SCALE:
+            match sizing_mode:
+                case Sizing.SMALLER:
+                    pass
+                case Sizing.BIGGER:
+                    image_copy.thumbnail((max_width, max_height), Image.LANCZOS)
+                case Sizing.TALLER:
+                    height_ratio = current_image_height / max_height
+                    new_width = int(current_image_width // height_ratio)
```
swfarnsworth, line 176: It's not necessary to wrap this in `int(  )` as floor division with `//` returns an int.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+                    image_copy.thumbnail((new_width, max_height), Image.LANCZOS)
+                case Sizing.WIDER:
+                    width_ratio = current_image_width / max_width
+                    new_height = int(current_image_height // width_ratio)
+                    image_copy.thumbnail((max_width, new_height), Image.LANCZOS)
+    return image_copy
+
+
+def image_size_compare(
+    image_width: int, image_height: int, max_width: int, max_height
```
swfarnsworth, line 186: Looks like you forgot the type hint for `max_height`--not that I was wondering what type it should be 😛

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    Removes non-ASCII characters from a string.
+
+    param text: String to convert.
+    :return: The string minus any non-ASCII characters.
+    """
+    # https://stackoverflow.com/questions/2758921/regular-expression-that-finds-and-replaces-non-ascii-characters-with-python
+    return text.encode().decode("ascii", "replace").replace("\ufffd", "")
+
+
+def to_bytes(text: str) -> list[int]:
```
swfarnsworth, line 217: It's a bit misleading to call this `to_bytes` considering that `bytes` is a data type that isn't `list[int]`.

thick-wrappers/helper/utility.py
```diff
@@ -0,0 +1,233 @@
...
+    return list(bytearray(text, "ascii"))
+
+
+def list_images(dir: str) -> list[str]:
+    """Lists all .png or .jpeg images in a directory
+
+    param dir: Directory to list.
+    :return: List of paths to images in directory.
+    """
+    return glob(join(dir, "*.png")) + glob(join(dir, "*.jp?g"))
```
swfarnsworth, line 233: Here's how this could be improved with pathlib, if you had `dir` as a path instead of a string.
```py
return list(dir.glob('*.png')) + list(dir.glob('*.jp?g'))
```
While this isn't any shorter, it doesn't require any additional imports, as it just uses `list` and the `Path` object's own glob method.

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+    def get_decrypted_output_file_path(self, text=False):
+        """Get file path for decrypted output, either an image or text file"""
+        if text:
+            return os.path.join("static", "output.txt")
+        else:
+            return os.path.join("static", f"output.{self.cover_image_fe}")
+
+
+@dataclass
+class TailwindStyling:
```
swfarnsworth, line 49: It looks like you're just using this class as a namespace, rather than as a basis for a custom data type. If you just want a namespace, use `types.SimpleNamespace`.

```py
tailwind_styling = SimpleNamespace(
    header_row = "flex items-center justify-center relative px-4"
    center_card = "absolute top-1/2 left-1/2 transform -translate-x-1/2 \
    -translate-y-1/2 p-8 rounded-lg shadow-md max-w-md"
    button_row = "flex"
    # etc.
)
```

You can then access each attribute with expressions like `tailwind_styling.header_row`, much like a dict but with attributes instead of keys.

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+                cover_image_upload_text.reset()
+            else:
+                cover_image_upload_image.reset()
+        else:
+            secret_image_upload.reset()
+        return
+    # Get the extension and check that it is present and valid
+    acceptable_extensions = ["jpg", "png", "jpeg"]
+    file_extension = os.path.splitext(img.name)[1]
+    if not file_extension or file_extension[1:] not in acceptable_extensions:
```
swfarnsworth, line 142: With pathlib, this could be written as one statement:
```py
if img.name.suffix not in [".jpg", ".png", ".jpeg"]:
```
note that the `suffix` of a Path will have a leading `.`, or will be an empty string (and thus not be in the list) if there isn't an extension.

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+    # Flush upload prompts
+    secret_image_upload.reset()
+    decrypt_image_upload.reset()
+    show_output()
+
+
+# GUI Contents
+
+# Create Filepaths object to keep track of file paths
+file_paths = Filepaths()
```
swfarnsworth, line 305: It looks like the functions above are reading this variable from the global scope. While you've written this in such a way that that won't cause any problems, it's a good practice to define global variables before any code that refers to them.

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+# Create TailwindStyling object for components styles
+styles = TailwindStyling()
+
+# Add static files folder
+app.add_static_files("/static", "static")
+# Add dark mode config
+dark_mode = ui.dark_mode()
```
swfarnsworth, line 312: Thank you for including dark mode 😄

thick-wrappers/main.py
```diff
@@ -0,0 +1,441 @@
...
+                    e,
+                    dropdown_text_or_image.value,
+                    enter_text_or_upload.value,
+                    (
+                        text_to_encrypt
+                        if isinstance(text_to_encrypt, str)
+                        else text_to_encrypt.value
+                    ),
+                ),
+            )
```
swfarnsworth, line 388: Nice formatting here

thick-wrappers/tests/test_image.py
swfarnsworth, line 1: I appreciate that you've written tests and that you're using pytest instead of unittest.

thick-wrappers/tests/test_text.py
```diff
@@ -0,0 +1,58 @@
+import string
+from random import choices, seed
+
+from PIL import Image
+
+from helper import decrypt, encrypt, utility
+
+seed(a=1)
+ascii = string.ascii_letters + string.digits + string.punctuation + "\n\t\r"
+non_ascii = "".join(map(chr, range(128, 1000)))
```
swfarnsworth, line 10: Tasteful use of `map`. (I dislike seeing `list(map(...))` or maps involving lambdas.)

swfarnsworth (commented):
Hey there, Thick Wrappers! Thank you for participating in this year's code jam, and congratulations on making it this far! I've enjoyed reviewing your submission, and I hope that my comments help you improve as Python developers.

I also hope that you'll take my suggestions for improvement in stride. I am overall impressed with what you came up with, especially among a team of strangers in so short a time.

I think the project you came up with is a fun idea, and a great application of image processing and the theme. I also appreciate that contributes are reasonably balanced among all the team members.

We'll see you at the live stream!

PR #10, opened 2023-09-28
Top Level Walruses

janine9vn:


top-level-walruses/piet/cli.py
```diff
@@ -0,0 +1,80 @@
...
+from .interpreter import PietInterpreter
+
+app = typer.Typer()
+
+
+@app.command()
+def generate(
+    input_path: Path,
+    output_path: Path,
+    cols: Optional[int] = None,
```
janine9vn, line 18: This is minor, but the *slightly* recommended method of this type hint would be `int | None`.

janine9vn, line 18: This is minor, but the *slightly* recommended way to do this typehint would be: `int | None`

top-level-walruses/piet/cli.py
```diff
@@ -0,0 +1,80 @@
...
+        key = first_input.encode()
+        if key:
+            key *= len(data) // len(key) + 1
+            command += f' --input "{first_input}"'
+        image = generator.generate_image(data, key, cols)
+    image.save(output_path)
+    print("\n")
+    print("Successfully generated a Piet program!")
+    print(f"Image saved to: {output_path}")
+    print(f"You can run it with: {command}")
```
janine9vn, line 46: This is a nice QoL feature for the user, that you show the command they can run.

top-level-walruses/piet/cli.py
```diff
@@ -0,0 +1,80 @@
...
+    generator = ImageGenerator(input=first_input, step_limit=step_limit, debug=debug)
+    if recurse:
+        image = generator.generate_recursive(data, (v.encode() for v in input), recurse)
+        command += " --execute"
+    else:
+        key = first_input.encode()
+        if key:
+            key *= len(data) // len(key) + 1
+            command += f' --input "{first_input}"'
+        image = generator.generate_image(data, key, cols)
```
janine9vn, line 41: I think this function could use a docstring or some comments in this area about what's happening. It's doing a fair bit and not all of it necessarily straightforward.

top-level-walruses/piet/common.py
```diff
@@ -0,0 +1,143 @@
+from copy import deepcopy
+from typing import Iterable, NamedTuple, TypeVar, overload
+
+T = TypeVar("T")
+
+
+class SelfExpandingList(list[T]):
```
janine9vn, line 7: This class is doing quite a bit. It is in desperate need of some explainer comments about what its purpose is and what it is trying to do.

You can piece together what it is doing and its purpose after a bit, but comments would go a long way.

top-level-walruses/piet/common.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class ColorChange(NamedTuple):
+    lightness: int
+    hue: int
+
+
+#         lightness hue
+#              v     v
+PIET_COLORS: tuple[tuple[Color, ...], ...] = (
```
janine9vn, line 116: Cute comment and very helpful

top-level-walruses/piet/generator.py
```diff
@@ -0,0 +1,221 @@
...
+        self.debug = debug
+        self.commands = SelfExpandingList(default=SelfExpandingList[PietCommand](default=PietCommand._NONE))
+        self.colors = SelfExpandingList(default=SelfExpandingList(default=BLACK))
+        self.interpreter = PietInterpreter(self.image, input=input, step_limit=step_limit, debug=debug)
+        self.interpreter.runtime.output = open(sys.stdout.fileno(), "wb", closefd=False)
+        self._current_hue = 0
+        self._current_lightness = 0
+        self._previous_color = WHITE
+
+    def set_command(self, command: PietCommand, position: OrderedPair):
```
janine9vn, line 58: A docstring would be very helpful here, particuarly to help explain some of the modulo math happening under the first if statement.

top-level-walruses/piet/generator.py
```diff
@@ -0,0 +1,221 @@
...
+            color = command.value
+        else:
+            color = WHITE
+        self.colors[y][x] = color
+        self._previous_color = color
+
+    def set_offset_command(self, command: PietCommand, offset: DirectionOffset):
+        pointer = DirectionPointer()
+        pointer.direction = self.interpreter.runtime.pointer.direction
+        pointer.position = self.interpreter.runtime.pointer.position
```
janine9vn, line 76: Why does this need separate pointers initialized with the current values of the interpreter? What is the purpose of this function? Why is an offset useful?

top-level-walruses/piet/cli.py
```diff
@@ -0,0 +1,80 @@
...
+    cols: Optional[int] = None,
+    recurse: int = 0,
+    input: list[str] = [],
+    step_limit: int = 1_000_000,
+    debug: bool = False,
+):
+    print("Generating a Piet program...")
+    print("This process may take a while, depending on the size of the data.")
+    if input_path.exists():
+        data = input_path.read_bytes()
```
janine9vn, line 27: This will be true for directories, not just files. When you try to read from a directory, it will fail. You should check if it's specifically a file or implement a try/except.

top-level-walruses/piet/common.py
```diff
@@ -0,0 +1,143 @@
...
+            if isinstance(index, slice):
+                items = []
+                for i in range(index.start or 0, index.stop or len(self), index.step or 1):
+                    try:
+                        items.append(super().__getitem__(i))
+                    except IndexError:
+                        items.append(deepcopy(self._default))
+                return self.__class__(items, default=deepcopy(self._default))
+            if index < -1:
+                raise IndexError("SelfExpandingList index out of range")
```
janine9vn, line 31: This check implies that `-1` is a valid index. Is that true for this implementation? It seems unlikely. Similar comment can be applied to the `__setitem__` implementation as well.

top-level-walruses/piet/runtime.py
```diff
@@ -0,0 +1,215 @@
...
+        pos = self.input.tell()
+        while True:
+            char = self.input.read(1)
+            if char in whitespace and buffer:
+                self.input.seek(pos + len(buffer))
+                break
+            buffer += char
+            if char in whitespace and not buffer:
+                self.input.seek(pos + len(buffer))
+                break
```
janine9vn, line 194: The `not buffer` check on line 192 is unnecessary as the line before will always push onto the buffer.

This section as a whole can potentially be re-written as the following to reduce duplicate functionality:
```py
            if char in whitespace:
                if not buffer:
                    buffer += char
                self.input.seek(pos + len(buffer))
                break
```

top-level-walruses/piet/reader.py
```diff
@@ -0,0 +1,105 @@
...
+        return Codel(color, visited)
+
+    def smallest_codel(self) -> int:
+        """Return the side length in pixels of the smallest codel."""
+        colors = self.colors
+        pixel = colors[0][0]
+        height, width = len(colors), len(colors[0])
+        smallest = height
+
+        rsf = 0
```
janine9vn, line 69: What is rsf? It is unclear in this section what it is intended to represent.

top-level-walruses/piet/generator.py
```diff
@@ -0,0 +1,221 @@
...
+    def image(self) -> Image.Image:
+        colors = self.colors
+        if not colors:
+            return Image.new("RGB", (1, 1), WHITE)
+        height = len(colors)
+        width = len(max(colors, key=len))
+        color_ints = []
+        for y in range(height):
+            for x in range(width):
+                color_ints.append(int(colors[y][x]))
```
janine9vn, line 134: This can be replaced with some functions from itertools:
```py
import itertools

...

for color in itertools.chain.from_iterable(colors):
  color_ints.append(int(color))
```
or, depending on what `width` is, just `itertools.product`

top-level-walruses/piet/reader.py
```diff
@@ -0,0 +1,105 @@
...
+                    pixel = colors[y][x]
+                    rsf = 0
+                next = colors[y][x]
+                if next == pixel:
+                    rsf += 1
+                else:
+                    if rsf < smallest:
+                        smallest = rsf
+                        pixel = colors[y][x]
+                        rsf = 1
```
janine9vn, line 82: This can be simplified into an `elif`

top-level-walruses/piet/interpreter.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"StepTrace({self.iteration:0>4},"
+            f" instruction={self.instruction.__name__},"
+            f" position={self.position},"
+            f" did_flip={self.did_flip},"
+            f" last_codel={self.last_codel},"
+            f" current_codel={self.current_codel})"
+        )
```
janine9vn, line 42: Very nice usage of a repr here!

top-level-walruses/piet/runtime.py
```diff
@@ -0,0 +1,215 @@
...
+from warnings import warn
+
+from .common import OrderedPair
+
+
+class PointerDirection(Enum):
+    RIGHT = OrderedPair(0, 1)
+    DOWN = OrderedPair(1, 0)
+    LEFT = OrderedPair(0, -1)
+    UP = OrderedPair(-1, 0)
```
janine9vn, line 16: This is a nice way of handling directions.

janine9vn (commented):
## Code Quality
The code itself is very well done. In particular some of the extra work gone into the files in `common.py` is appreciated, to clearly define non-standard behavior and avoid messy hacks that would otherwise be necessary. The typehints are well done and consistent as well. The code is logical and (mostly) easy to follow and reason about. 

## Documentation
One aspect that was lacking in this project was documentation. Many functions did not have docstrings, and when they did, they often felt lacking. Some of the more dense code as well could have substantially benefitted from comments explaining what the intent is and some of the more technical details. I did not comment on every instance, but it was pretty much project-wide and did make this project a bit harder to review.

The README/presentation was well done. I appreciate the demo examples included and how everything was explained. The one aspect I think could be improved in general for the README, or in the code itself, is a more robust explanation of the different components of the Piet interpreter, generator, reader, runtime, and the common files. There's a lot going on there and could be benefitted from more words to help explain what's happening.

## Commit Messages
The commit messages overall were good. Some could have been a bit more descriptive about what the commit itself accomplished. For instance "piet interpreter almost finished", could instead describe what was added to get the piet interpreter to an almost finished state. In general, for more substantial commits, I'm a fan of being more specific about what was added in the commit body. 

## Summary
This is a very impressive project, with a fully working interpreter *and* generator for Piet. The lack of documentation is a sore spot, but otherwise I don't have major comments. The organization and structure of the project as well is nicely done and logical. It made it easier to review. Additionally, I do appreciate the inclusion of tests. This code jam is usually a time crunch, so including tests is going a bit above and beyond.

### A quick note...
Thank you to dawn and aboo for reviewing parts of this project and providing some of the comments I left.

top-level-walruses/piet/cli.py
lemonyte: IIRC `Optional` is here because Typer does not support the new-style annotation `type | None`.

top-level-walruses/piet/common.py
lemonyte: To be honest, this was more of a last-ditch attempt to facilitate a dynamically sized 2-dimentional array without lengthy blocks of logic and `append` calls.

top-level-walruses/piet/generator.py
lemonyte: Similar to the `set_command` function, this function needs to know the current position of the internal interpreter, so it can pre-determine the next location of the pointer and place the command where it will be executed next. It needs a temporary pointer to simulate moving to the offset position before placing the command. The offset is useful for modifying commands not in the next position of the pointer, but offset in one of 8 directions (N, E, S, W, NE, NW, SE, SW). For example, placing black codels ahead of the next position in order to force a specific execution path.

